PR #1, opened 2020-06-15
initial import

nden:
@eslavich @perrygreenfield @jdavies-st

This is the initial import of the transform tags. This includes a new extension registered with asdf through entry points. I kept some of the old code here (commented out) as in the future other tags will be moved, so we'll simply uncomment those lines.
As for the new extension if this is not the way to go, either let me know how to fix it or we can merge and improve.  This by itself is not sufficient to write out a file using the new schemas and tags.


perrygreenfield:
Which code is commented out? I guess there are two strategies here possible. Disable all tag code and get each working piece by piece, and uncomment the corresponding tests. Or assume a mixed mode that transitions from the old tag style to the new converter style. I think I would prefer the latter (it makes it easier to see what isn't done yet, but it means it all has to get done before it can be used in general). But  I may be confused.

nden:
I am adding the new Converter to this PR. Will push shortly.

asdf_astropy/tags/helpers.py
```diff
@@ -0,0 +1,22 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+
+import numpy as np
+
+
+__all__ = []
+
+
+def skycoord_equal(sc1, sc2):
```
jdavies-st, line 10: I believe `SkyCoord` now has `__eq__` defined for strict equality.

asdf_astropy/tags/tests/helpers.py
```diff
@@ -0,0 +1,36 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# -*- coding: utf-8 -*-
+import numpy as np
+
+
+def run_schema_example_test(organization, standard, name, version, check_func=None):
```
jdavies-st, line 6: Does this do the same thing as the pytest asdf schema tester plugin?  It also runs examples.  Is this doing anything different?

asdf_astropy/tests/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+from pkg_resources import iter_entry_points
+
+entry_points = [entry.name for entry in iter_entry_points('asdf_extensions')]
+ASDF_ENTRY_INSTALLED = 'astropy' in entry_points and 'asdf-astropy' in entry_points
+
+del entry_points, iter_entry_points
+
+if not ASDF_ENTRY_INSTALLED:
+    pytest.skip('The astropy asdf entry points are not installed',
+                allow_module_level=True)
```
jdavies-st, line 17: We may not need any of this, as this is the package that will be providing the `asdf_extensions` entry points, so there's no check that needs to be done.

asdf_astropy/tests/test_example.py
```diff
@@ -0,0 +1,13 @@
...
+    from ..example_mod import primes
+    assert primes(10) == [2, 3, 5, 7, 11, 13, 17, 19, 23, 29]
+
+
+def test_deprecation():
+    import warnings
+    warnings.warn(
+        "This is deprecated, but shouldn't raise an exception, unless "
+        "enable_deprecations_as_exceptions() called from conftest.py",
+        DeprecationWarning)
```
jdavies-st, line 13: These can probably be deleted?  Perhaps the whole module?

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+
+requires = ["setuptools",
+            "setuptools_scm",
+            "wheel",
+            #"extension-helpers",
+            "oldest-supported-numpy",
```
jdavies-st, line 7: You need a closing `]` here.

setup.cfg
```diff
@@ -0,0 +1,76 @@
+[metadata]
+name = asdf_astropy
+author = The ASDF dev team
+author_email =
```
jdavies-st, line 4: email needed.

.travis.yml
```diff
@@ -0,0 +1,161 @@
...
+        # The latter can be used for example to pass arguments to pytest.
+        - TOXENV='test'
+        - TOXARGS='-v'
+        - TOXPOSARGS=''
+
+        # The following is needed to avoid issues if e.g. Matplotlib tries
+        # to open a GUI window.
+        - SETUP_XVFB=True
+
+matrix:
```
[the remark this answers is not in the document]
nden: It came from the package-template, this should be mentioned there.

nden:
I am going to close this PR and open a new one which uses the converter classes. Sorry for making you review it.

PR #2, opened 2020-06-16
Initial import of the package

nden:
This is the initial import of the package working with the new converters. The new extension (provider) is set up, as well as the converters in basic.py and polynomial.py. Works for me locally - example with Chebyshev1D:

```
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
  name: asdf, version: 2.6.1.dev59+gde25fec.d20200615}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension.BuiltinExtension
    software: !core/software-1.0.0 {name: asdf, version: 2.6.1.dev59+gde25fec.d20200615}
m: !<http://asdf-format.org/schemas/transform/ortho_polynomial-2.0.0>
  coefficients: !core/ndarray-1.0.0
    data: [0.0, 0.0]
    datatype: float64
    shape: [2]
  polynomial_type: chebyshev
  window: [-1, 1]
...

```


asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+from astropy.modeling.core import CompoundModel, Model
+
+from . import _parameter_to_value
+
+
+__all__ = ['TransformType', 'IdentityType', 'ConstantType']
+
+
+class TransformType(asdf.AsdfConverter):
+    tags = {"http://asdf-format.org/schemas/transform/transforms-1.0.0"}
```
eslavich, line 17: I would just leave out `tags` and `types` here, since we don't want to actually write a generic transform to a file.

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+    types = {Model}
+    version = '2.0.0'
```
eslavich, line 19: The `version` and `requires` attributes are no longer recognized by the framework

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+        return node
+
+    def to_yaml_tree(self, model):
+        raise NotImplementedError("Must be implemented in TransformType subclasses")
+
+    def to_tree(self, model):
+        node = self.to_tree_transform(model)
+        return self._to_tree_base_transform_members(model, node)
+
+    def assert_equal(self, a, b):
```
eslavich, line 100: These `assert_equal` methods are something that I haven't given much thought yet.  It would be nice to put the responsibility of defining equality onto astropy itself, where upkeep will be easier.

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+        # TODO: If models become comparable themselves, remove this.
+        assert a.name == b.name
+        # TODO: Assert inverses are the same
+
+
+class IdentityType(TransformType):
+    tags = {"http://asdf-format.org/schemas/transform/identity-2.0.0"}
```
eslavich, line 107: We'll still need to support the old tags, so this list should also include:

```python
"tag:stsci.edu:asdf/transform/identity-1.0.0",
"tag:stsci.edu:asdf/transform/identity-1.1.0",
"tag:stsci.edu:asdf/transform/identity-1.2.0",
```
For transforms whose schemas see a lot of changes in this migration, we may want to maintain legacy support in a separate class, just so things don't get confusing.

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+        # TODO: If models become comparable themselves, remove this.
+        TransformType.assert_equal(a, b)
+        assert (isinstance(a, models.Identity) and
+                isinstance(b, models.Identity) and
+                a.n_inputs == b.n_inputs)
+
+
+class ConstantType(TransformType):
+    tags = {"http://asdf-format.org/schemas/transform/constant-2.0.0"}
+    supported_versions = ['1.0.0', '1.1.0', '1.2.0', '1.3.0', '1.4.0', '2.0.0']
```
eslavich, line 129: `supported_versions` is also now ignored, that is determined by the list of tags.

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+    types = {models.Const1D, models.Const2D}
+
+    def from_yaml_tree(self, node):
+        if self.version < AsdfVersion('1.4.0'):
```
eslavich, line 133: The `self.version` property still exists, though at the moment it returns a string, I should probably change that.

eslavich: (I'm not a fan of `AsdfVersion` and hope to deprecate it)

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,216 @@
...
+            elif isinstance(model, models.Const2D):
+                dimension = 2
+            return {
+                'value': _parameter_to_value(model.amplitude),
+                'dimensions': dimension
+            }
+
+
+class UnitsMappingType(asdf.AsdfConverter):
+    tags = {"http://asdf-format.org/schemas/transform/units_mapping-2.0.0"}
```
eslavich, line 162: This transform schema was deemed to be too astropy-specific for the general use transforms, so it's actually still living in astropy with URI `http://astropy.org/schemas/astropy/transform/units_mapping-1.0.0`.

asdf_astropy/tags/transform/polynomial.py
```diff
@@ -0,0 +1,301 @@
...
+                assert a.domain == b.domain
+                assert a.window == b.window
+            else:
+                assert a.x_domain == b.x_domain
+                assert a.x_window == b.x_window
+                assert a.y_domain == b.y_domain
+                assert a.y_window == b.y_window
+
+
+class PolynomialType1_0(PolynomialTypeBase):
```
eslavich, line 179: Ha!  I forgot about these.  I did this because I didn't understand how supported_versions worked in the old framework.  We can merge these back into one class again and just list multiple `tags`.

asdf_astropy/extension.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+__all__ = ['TransformConverterProvider']
+
+
+class TransformConverterProvider(asdf.AsdfConverterProvider):
+    converter_classes = list(_asdf_format_types)
+
+
+asdf.configure(converter_providers=[TransformConverterProvider()])
```
eslavich, line 50: I don't think we want to call `configure` from within a library, since that will override other providers that the user might have installed.  The entry point should take care of this for us (but maybe it's broken right now, I can't remember if I tested that specifically).

asdf_astropy/tags/transform/basic.py
nden: Agreed!

asdf_astropy/tags/transform/basic.py
nden: Can a third party implementation of the standard implement only part of the schemas? If so, perhaps we should keep all schemas in one place and let people decide what they want to implement. I think it's something we should discuss further.

eslavich: That sounds like a good idea to me, there are so many transforms that the prospect of implementing them all is intimidating.

I wonder if we should separate them into namespaces to guide people on implementing a minimal set?  `transform/core` and others.

asdf_astropy/extension.py
nden: A quick test shows it doesn't work without it.

eslavich: Ah okay thanks, I'll investigate

setup.cfg
```diff
@@ -0,0 +1,76 @@
...
+python_requires = >=3.6
+setup_requires = setuptools_scm
+install_requires =
+    astropy
+    asdf
+    numpy
+    asdf-transform-schemas
+
+[options.entry_points]
+asdf_extensions =
```
eslavich, line 26: The name of the entry point is different for the new-style plugin.  This should be:

```
asdf.converter_providers =
   asdf-astropy = asdf_astropy.extension:TransformConverterProvider
```

asdf_astropy/extension.py
eslavich: I think I found the issue, left a comment here:

https://github.com/astropy/asdf-astropy/pull/2/files#r441786043

asdf_astropy/tags/transform/basic.py
perrygreenfield: This worries me as well. I don't think people should be forced to implement them all (e.g., all the projection transform)

asdf_astropy/tags/transform/basic.py
```diff
@@ -0,0 +1,223 @@
...
+from astropy.modeling import models
+from astropy.modeling.core import CompoundModel
+
+from . import _parameter_to_value
+
+
+__all__ = ['TransformType', 'IdentityType', 'ConstantType']
+
+
+class TransformType(asdf.AsdfConverter):
```
perrygreenfield, line 16: Instead change the class names to use Converter instead of Type?

asdf_astropy/tags/transform/compound.py
```diff
@@ -0,0 +1,166 @@
...
+            "tag:stsci.edu:asdf/transform/compose-1.1.0",
+            "tag:stsci.edu:asdf/transform/compose-1.2.0",
+            "tag:stsci.edu:asdf/transform/concatenate-1.0.0",
+            "tag:stsci.edu:asdf/transform/concatenate-1.1.0",
+            "tag:stsci.edu:asdf/transform/concatenate-1.2.0",
+            "tag:stsci.edu:asdf/transform/fix_inputs-1.1.0",
+            "tag:stsci.edu:asdf/transform/fix_inputs-1.2.0",
+            }
+    types = {CompoundModel}
+    handle_dynamic_subclasses = True
```
eslavich, line 71: This attribute controls a feature that we're planning to remove (and not implement in `AsdfConverter`), so we can drop it.

PR #9, opened 2021-01-16
fix tox envs

nden:
Updated the tox environemnts. The tests run now locally.
One test with the astropy LTS release fails. I suspect a commit which fixed it did not propagate to the release but we need to track this down and if so include it in the next LTS patch.

### DIFF
--- a/asdf_astropy/converters/transform/tabular.py
+++ b/asdf_astropy/converters/transform/tabular.py
@@ -17,7 +17,8 @@ class TabularConverter(TransformConverterBase):
 
     def to_yaml_tree_transform(self, model, tag, ctx):
         node = {}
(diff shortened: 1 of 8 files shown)


asdf_astropy/converters/transform/tabular.py
```diff
@@ -17,7 +17,8 @@ class TabularConverter(TransformConverterBase):
 
     def to_yaml_tree_transform(self, model, tag, ctx):
         node = {}
-        node["fill_value"] = model.fill_value
+        if model.fill_value is not None:
```
eslavich, line 20: A victim of the null value change?

eslavich (approved):
LGTM

Issue #12, opened 2021-03-09
MNT: Rename default branch from master to main


pllim:
Please rename your default branch from `master` to `main` as part of astropy/astropy-project#151 , preferably by 2021-03-22. Also a friendly reminder to check documentation, workflows, etc., and update them accordingly. Please don't forget to communicate this change to your users and stakeholders. To summarize:

- [x] Rename branch from `master` to `main`, preferably by 2021-03-22.
- [x] Update documentation, workflows, etc., accordingly. -- See #13 
- [x] Communicate this change to your users and stakeholders.

Once this is taken care of, you may close this issue.

*This is an automated issue. If this is opened in error, please let @pllim know!*

eslavich:
@nden I was able to push a new `main` branch, but I don't appear to have the permissions necessary to change the default branch on the repo.  Are you able to do that?

pllim:
Actually, you could have just renamed it from admin settings...

pllim:
I don't have the permissions either. You might have to contact @astropy/coordinators .

kelle:
I just switched the default branch to the new `main` branch. If the settings of this repo need to be updated, please open a new issue.

eslavich:
Splendid, thanks!

eslavich:
This package hasn't been released yet, and I've notified the developers, so I think we're all set here.

PR #13, opened 2021-03-10
MNT: Replace master with main

pllim:
This PR attempts to replace all mention of `master` to `main` in this repository. This should be reviewed and merged right after the maintainer has renamed the default branch.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

This is part of #12

### DIFF
--- a/.github/workflows/asdf_astropy_ci.yml
+++ b/.github/workflows/asdf_astropy_ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
     tags:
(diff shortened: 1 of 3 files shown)


pllim:
I think this was lost in the renaming...  I don't see it in the `main` commit history. You might need to manually port this over. FYI.

eslavich:
Oh oops, another consequence of me creating a separate `main` branch instead of renaming `master` 🤦 .  Thanks, I'll make sure this gets in there.

PR #19, opened 2021-05-13
MNT: Simplified _astropy_init.py more

pllim:
Follow up of #17 

After I opened that PR, I was told more could be removed in another channel, so here it is.


eslavich (approved):
Even better!

PR #20, opened 2021-05-17
Add converter support for unit-related tags

eslavich:
This PR adds converter support for the following tags:

```
tag:astropy.org:astropy/units/equivalency-1.0.0
tag:stsci.edu:asdf/unit/quantity-1.1.0
tag:stsci.edu:asdf/unit/unit-1.0.0
```

### DIFF
--- a/asdf_astropy/converters/unit/equivalency.py
+++ b/asdf_astropy/converters/unit/equivalency.py
@@ -21,4 +21,9 @@ class EquivalencyConverter(Converter):
             kwargs = dict(zip(equivalency_node["kwargs_names"], equivalency_node["kwargs_values"]))
             components.append(equivalency_method(**kwargs))
 
+        # The Equivalency class is a UserList that overrides __add__ to
+        # provide special behavior when combined with another Equivalency.
+        # We're using sum here to add each subsequent Equivalency to the
+        # first so we end up with a single correctly combined Equivalency
+        # object at the end.
         return sum(components[1:], components[0])
--- a/asdf_astropy/converters/unit/equivalency.py
+++ b/asdf_astropy/converters/unit/equivalency.py
@@ -1,0 +1,36 @@
+from asdf.extension import Converter
+
+
+class EquivalencyConverter(Converter):
+    tags = ["tag:astropy.org:astropy/units/equivalency-*"]
+
+    types = ["astropy.units.equivalencies.Equivalency"]
+
+    def to_yaml_tree(self, obj, tag, ctx):
+        results = []
+
+        for name, kwargs in zip(obj.name, obj.kwargs):
+            kwargs_names = []
+            kwargs_values = []
+            for kwargs_name, kwargs_value in kwargs.items():
+                kwargs_names.append(kwargs_name)
+                kwargs_values.append(kwargs_value)
+
+            results.append({
+                "name": name,
+                "kwargs_names": kwargs_names,
+                "kwargs_values": kwargs_values,
+            })
+
+        return results
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from astropy.units import equivalencies
+
+        components = []
+        for equivalency_node in node:
+            equivalency_method = getattr(equivalencies, equivalency_node["name"])
+            kwargs = dict(zip(equivalency_node["kwargs_names"], equivalency_node["kwargs_values"]))
+            components.append(equivalency_method(**kwargs))
+
+        return sum(components[1:], components[0])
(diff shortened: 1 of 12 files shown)


asdf_astropy/converters/unit/equivalency.py
```diff
@@ -0,0 +1,36 @@
...
+            kwargs_values = []
+            for kwargs_name, kwargs_value in kwargs.items():
+                kwargs_names.append(kwargs_name)
+                kwargs_values.append(kwargs_value)
+
+            results.append({
+                "name": name,
+                "kwargs_names": kwargs_names,
+                "kwargs_values": kwargs_values,
+            })
```
mcara, line 23:
```suggestion
        results = [
            {'name': name, 'kwargs_names': list(kw.keys()), 'kwargs_values': list(kw.values())}
            for name, kw in zip(obj.name, obj.kwargs)
        ]
```

asdf_astropy/converters/unit/equivalency.py
```diff
@@ -0,0 +1,36 @@
...
+
+        return results
```
mcara, line 25:
```suggestion
        return [
            {'name': name, 'kwargs_names': list(kw.keys()), 'kwargs_values': list(kw.values())}
            for name, kw in zip(obj.name, obj.kwargs)
        ]
```

eslavich: Are `keys()` and `values()` guaranteed to return items in the same order when called independently?  I assumed not, but I don't know one way or the other...

eslavich: Oh, right, yes, dicts are ordered now.  Thanks, this is much tidier.

mcara: @perrygreenfield What do you think?

mcara: From https://docs.python.org/3/library/stdtypes.html#dict-views:

> The objects returned by dict.keys(), dict.values() and dict.items() are view objects. They provide a dynamic view on the dictionary’s entries, which means that when the dictionary changes, the view reflects these changes.

and

> Keys and values are iterated over in insertion order. This allows the creation of (value, key) pairs using zip(): pairs = zip(d.values(), d.keys()).

asdf_astropy/converters/unit/equivalency.py
```diff
@@ -0,0 +1,24 @@
...
+    def from_yaml_tree(self, node, tag, ctx):
+        from astropy.units import equivalencies
+
+        components = []
+        for equivalency_node in node:
+            equivalency_method = getattr(equivalencies, equivalency_node["name"])
+            kwargs = dict(zip(equivalency_node["kwargs_names"], equivalency_node["kwargs_values"]))
+            components.append(equivalency_method(**kwargs))
+
+        return sum(components[1:], components[0])
```
jdavies-st, line 24: What does the `sum` do here?  A comment would be nice.

asdf_astropy/converters/unit/tests/test_unit.py
```diff
@@ -0,0 +1,55 @@
...
+
+@pytest.fixture(autouse=True)
+def remove_astropy_extensions():
+    """
+    Disable the old astropy extension so that it doesn't
+    confuse our test results.
+    """
+    with asdf.config_context() as config:
+        config.remove_extension(package="astropy")
+        yield
```
jdavies-st, line 20: Since this gets used 3 times, you may wanat to put this in a `conftest.py`, turn off `autouse`, and make the fixture dependency explicit in each test that uses it.

jdavies-st (approved):
Looks good!  Some minor comments below.

asdf_astropy/converters/unit/equivalency.py
eslavich: It concatenates the lists together -- it's done this way to make sure we get an Equivalency object at the end instead of a vanilla list.  I'll add a comment.

asdf_astropy/converters/unit/tests/test_unit.py
eslavich: In the end this is going to be `autouse=True` for the entire package, but we can't do that until all the converters are implemented.  I thought it'd be easier to go through and delete all these fixtures vs delete the fixture name from individual test method signatures.

jdavies-st: Ah, I see.  Carry on.  👍

perrygreenfield (approved):
LGTM, especially with the new commentary on the sum of all equivalencies, soon coming to a movie theater near you.

PR #23, opened 2021-05-24
Add converter support for table and fits tags

eslavich:
This PR adds support for the following tags:

```
tag:stsci.edu:asdf/core/column-1.0.0
tag:stsci.edu:asdf/core/table-1.0.0
tag:stsci.edu:asdf/fits/fits-1.0.0
tag:astropy.org:astropy/table/table-1.0.0
tag:astropy.org:astropy/fits/fits-1.0.0
```

### DIFF
--- a/asdf_astropy/converters/fits/tests/test_fits.py
+++ b/asdf_astropy/converters/fits/tests/test_fits.py
@@ -1,0 +1,105 @@
+import asdf
+from asdf.tests.helpers import yaml_to_asdf
+from astropy.io import fits
+import numpy as np
+from numpy.testing import assert_array_equal
+import pytest
+
+from asdf_astropy.testing.helpers import assert_hdu_list_equal
+
+
+HDUL = fits.HDUList()
+HEADER = fits.Header([
+    ("FOO", "BAR", "BAZ"),
+    ("SOMENUM", "11.0"),
+    ("EMPTY",)
+])
+HDUL.append(fits.PrimaryHDU(header=HEADER))
+HDUL.append(fits.ImageHDU(data=np.arange(100)))
+
+HDUL_WITH_TABLE = fits.HDUList()
+HDUL_WITH_TABLE.append(
+    fits.BinTableHDU.from_columns(
+        np.array([(0, 1), (2, 3)], dtype=[("A", int), ("B", int)])
+    )
+)
+
+TEST_HDULS = [
+    HDUL,
+    HDUL_WITH_TABLE,
+]
+
+
+@pytest.mark.parametrize("hdul", TEST_HDULS)
+def test_serialization(hdul, tmp_path):
+    file_path = tmp_path / "test.asdf"
+    with asdf.AsdfFile() as af:
+        af["hdul"] = hdul
+        af.write_to(file_path)
+
+    with asdf.open(file_path) as af:
+        assert_hdu_list_equal(af["hdul"], hdul)
+
+
+def test_asdf_tag():
+    yaml = """
+hdul: !fits/fits-1.0.0
+  - header:
+      - [SIMPLE, true, conforms to FITS standard]
+      - [BITPIX, 8, array data type]
+      - [NAXIS, 0, number of array dimensions]
+      - [EXTEND, true]
+      - []
+      - ['', Top Level MIRI Metadata]
+      - []
+      - [DATE, '2013-08-30T10:49:55.070373', The date this file was created (UTC)]
+      - [FILENAME, MiriDarkReferenceModel_test.fits, The name of the file]
+      - [TELESCOP, JWST, The telescope used to acquire the data]
+      - []
+      - ['', Information about the observation]
+      - []
+      - [DATE-OBS, '2013-08-30T10:49:55.000000', The date the observation was made (UTC)]
+  - data: !core/ndarray-1.0.0
+      data: [2, 3, 3, 4]
+      datatype: float32
+      shape: [4]
+    header:
+      - [XTENSION, IMAGE, Image extension]
+      - [BITPIX, -32, array data type]
+      - [NAXIS, 4, number of array dimensions]
+      - [NAXIS1, 4]
+      - [NAXIS2, 3]
+      - [NAXIS3, 3]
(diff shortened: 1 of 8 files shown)


asdf_astropy/converters/table/table.py
```diff
@@ -0,0 +1,101 @@
...
+        from astropy.table import Column, MaskedColumn
+        from numpy.ma.core import MaskedArray
+
+        data = node["data"]
+        if isinstance(data, NDArrayType):
+            # TODO: Why doesn't NDArrayType work?  This needs some research
+            # and documentation.
+            data = data._make_array()
+
+        if isinstance(data, MaskedArray):
```
eslavich, line 40: This check is missing in astropy and as a result the mask is lost on read.  Do we need to fix that bug in astropy too?

asdf_astropy/converters/fits/tests/test_fits.py
```diff
@@ -0,0 +1,105 @@
...
+HDUL_WITH_TABLE.append(
+    fits.BinTableHDU.from_columns(
+        np.array([(0, 1), (2, 3)], dtype=[("A", int), ("B", int)])
+    )
+)
+
+TEST_HDULS = [
+    HDUL,
+    HDUL_WITH_TABLE,
+]
```
jdavies-st, line 30: This defines a lot of global variables.  Could this instead be something like:

```
def _generate_test_hdulists():
    hdul = fits.HDUList()
    ...


@pytest.mark.parametrize("hdul", _generate_test_hdulists())
def test_serialization(hdul, tmp_path):
    ...
```

asdf_astropy/converters/table/tests/test_table.py
```diff
@@ -0,0 +1,279 @@
...
+    assert_time_equal,
+    assert_time_delta_equal,
+)
+
+
+def assert_description_equal(a, b):
+    if a in ("", None) and b in ("", None):
+        return
+
+    assert a == b
```
jdavies-st, line 23: Since this is a pytest assertion helper, you may want to include

```
__tracebackhide__ = True
```

in this function (and the assertion helpers below) so that when the assertion fails, it reports and rewrites the failure instead of regurgitating the code from the assertion helper.  This makes debugging failing tests much cleaner.

asdf_astropy/testing/helpers.py
```diff
@@ -66,3 +66,12 @@ def assert_time_delta_equal(a, b):
...
     assert_array_equal(a.sec, b.sec)
+
+
+def assert_hdu_list_equal(a, b):
+    assert len(a) == len(b)
+    for hdu_a, hdu_b in zip(a, b):
+        assert_array_equal(hdu_a.data, hdu_b.data)
+        assert len(hdu_a.header.cards) == len(hdu_b.header.cards)
+        for card_a, card_b in zip(hdu_a.header.cards, hdu_b.header.cards):
+            assert tuple(card_a) == tuple(card_b)
```
jdavies-st, line 77:
```
__tracebackhide__ = True
```

here too.

asdf_astropy/converters/table/tests/test_table.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def assert_table_equal(a, b):
+    assert type(a) == type(b)
+    assert a.meta == b.meta
+
```
jdavies-st, line 29:
```
__tracebackhide__ = True
```

asdf_astropy/converters/table/tests/test_table.py
```diff
@@ -0,0 +1,279 @@
...
+
+def assert_table_roundtrip(table, tmp_path):
+    """
+    Assert that a table can be written to an ASDF file and read back
+    in without losing any of its essential properties.
+    """
+    file_path = tmp_path / "testable.asdf"
+
+    with asdf.AsdfFile({"table": table}) as af:
```
jdavies-st, line 56:
```
__tracebackhide__ = True
```

jdavies-st (approved):
Looks good!  Just some minor testing comments below.

asdf_astropy/converters/fits/tests/test_fits.py
eslavich: Hmm... your suggestion also has the benefit of creating new objects for each test which better isolates them.  I used these global variables in all the test files, I'll make the suggested change everywhere.

eslavich: ... but how about in a separate PR, so reviewers of this one don't have to wade through it all

asdf_astropy/converters/table/tests/test_table.py
eslavich: I had `__tracebackhide__` originally, but since these helpers include multiple assertions, I found that I was missing useful detail on which assertion failed.

asdf_astropy/converters/fits/tests/test_fits.py
jdavies-st: No worries.  👍

asdf_astropy/extensions.py
```diff
@@ -443,14 +449,17 @@
     converters=ASTROPY_CONVERTERS,
 )
 
-# unit and quantity are part of the ASDF Standard core tags,
+# These tags are part of the ASDF Standard,
```
perrygreenfield, line 452: I regret allowing FITS stuff in the basic standard. Sorry for the digression. Maybe we can sneak it out one day.

perrygreenfield (approved):
No comments to add to those of @jdavies-st

asdf_astropy/extensions.py
eslavich: I wonder if anyone even needs to store FITS in the tree in a structured way.  Maybe the idea to allow embedding general files in blocks would serve FITS just as well (see email chain "Embedding general files in ASDF" from 2021-01-28).

asdf_astropy/converters/fits/tests/test_fits.py
eslavich: Here's the separate PR: https://github.com/astropy/asdf-astropy/pull/25

PR #29, opened 2021-11-03
Catch up to changes in ASDF support in astropy 5.0

eslavich:
This PR incorporates the following work done on astropy:

https://github.com/astropy/astropy/pull/12381
https://github.com/astropy/astropy/pull/11942
https://github.com/astropy/astropy/pull/11978

Also added an issue to support new models in astropy 5 and added the to the known unsupported models list so that the tests can pass.


nden (approved):
LGTM

PR #33, opened 2021-11-24
Use release version of asdf-transform-schemas

eslavich:
The asdf-transform-schemas package has been released, so we no longer need to install it from GitHub.  I'm also removing a test that has been made irrelevant by the change to asdf-transform-schemas to remove the ASDF Standard version requirements.

### DIFF
--- a/asdf_astropy/converters/transform/tests/test_transform.py
+++ b/asdf_astropy/converters/transform/tests/test_transform.py
@@ -402,6 +402,16 @@ def test_all_models_supported():
         assert len(missing) == 0, message
 
 
+@pytest.mark.xfail(reason="Older tag versions are difficult to test until asdf implements new config features")
+def test_legacy_const(tmpdir):
+    model = astropy_models.Const1D(amplitude=5.)
+    assert_model_roundtrip(model, tmpdir, version="1.3.0")
+
+    model = astropy_models.Const2D(amplitude=5.)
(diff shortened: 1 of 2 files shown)


asdf_astropy/converters/transform/tests/test_transform.py
```diff
@@ -402,15 +402,6 @@ def _iterate_model_classes():
...
 
 
-def test_legacy_const(tmpdir):
-    model = astropy_models.Const1D(amplitude=5.)
-    assert_model_roundtrip(model, tmpdir, version="1.3.0")
-
-    model = astropy_models.Const2D(amplitude=5.)
-    with pytest.raises(TypeError, match="does not support models with > 1 dimension"):
-        assert_model_roundtrip(model, tmpdir, version="1.3.0")
-
```
nden, line 412: Why is this test removed? Is it covered by another test?

eslavich: Now that I've removed the ASDF Standard version requirement from the transforms, asdf will use the latest version of the transform tags to write the file, which supports models with > 1 dimension.  It's not very easy to force asdf to use an earlier version of the transforms, which is what we'd need to do to test this error.  I suspect we'll need to add a new feature to the asdf config to support that.

eslavich: ... I should xfail this, shouldn't I.  I'll make that change in a sec.

PR #39, opened 2022-01-27
Bugfix for `units_mapping` asdf schema.

WilliamJamieson:
In asdf-format/asdf-transform-schemas#33, the `inputs` and `outputs` properties were formally added to the transform schema. This conflicts with the use of those fields in the `units_mapping` schema for other purposes. This PR fixes the bugs that result from this.

Note that this replicates the bugfix in PR astropy/astropy#12800.


eslavich (approved):
LGTM

PR #42, opened 2022-02-25
Add new trigonometric model converters

WilliamJamieson:
Add converters corresponding to those introduced in asdf-format/asdf-transform-schemas#40.


eslavich (approved):
Nice and tidy!

PR #43, opened 2022-02-25
Add a spline1D converter

WilliamJamieson:
Adds a converter for the schema introduced in asdf-format/asdf-transform-schemas#41.


eslavich (approved):
LGTM

PR #46, opened 2022-03-03
Remove python 3.7 from CI

WilliamJamieson:
Astropy 5 only supports python 3.8+ which means using new astropy features will break the Python 3.7 CI. Hence, we are dropping it here.


eslavich (approved):
LGTM

PR #47, opened 2022-03-07
Add astropy Table connector for ASDF

eslavich:
This copies in the Table connector for ASDF which is currently missing from this package.  I've tested it with existing releases of astropy as well as a dev branch that imports the connector from here.


WilliamJamieson:
@eteq can you confirm that the license related changes in this PR are correct?

WilliamJamieson (changes requested):
Everything looks good, this seems like its just a transfer of the connect code from astropy to this package.

Please fix the codestyle error, I think `bandit.yaml` needs to be modified to ignore the new test directory.

WilliamJamieson (approved):

eteq (approved):
The license stuff looks right to me.  The one caveat is: while my understanding is that this is a bit of a legal grey zone, I think it's probably good to specify just what is covered under the Astropy license.  So e.g., you just either put in the `ASTROPY_LICENSE` file something like "the  files xxx and yyy are included under this license", or alternatively just have the single-line license bit on the top of the relevant files but have the copyright holder be Astropy developers (like the core package does - see e.g. https://github.com/astropy/astropy/blob/main/astropy/convolution/core.py#L1 ).

WilliamJamieson (approved):

PR #52, opened 2022-03-08
Fix warnings from testing

WilliamJamieson:
This fixes the warnings occurring during testing and adds a CI job which turns warnings into exceptions to catch these issues in the future.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -96,3 +96,12 @@ deps=
     toml
 commands=
     bandit -c bandit.yaml -r .
+
+[testenv:warnings]
+commands=
+    pip freeze
+    pytest --remote-data -W error \
+      -p no:unraisableexception \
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -96,3 +96,12 @@ deps=
     toml
 commands=
     bandit -c bandit.yaml -r .
+
+[testenv:warnings]
```
pllim, line 100: Do you need a special job for this? Why not set this global for pytest in all the jobs?

WilliamJamieson: We don't like new warnings killing all of our CI. This one services as a friendly reminder to fix warnings without causing unrelated failures for normal development and testing.

pllim: But depending on how the matrix is set up, some warning might appear in one job but not another.

eslavich (approved):
LGTM

PR #57, opened 2022-03-24
Remove test skips for units_mapping schemas

WilliamJamieson:


eslavich (commented):
Should we make this stance official and update setup.cfg to require `astropy>=5`?

perrygreenfield (approved):
LGTM

PR #81, opened 2022-05-24
Add support for multiple tags in a single FrameConverter

### LINKED ISSUE #64: Add support for multiple tags in a single FrameConverter
Don't assume that `tag` is always a string.

WilliamJamieson:
Fixes #64


Cadair (approved):

Cadair:
Yep I think so

Issue #85, opened 2022-05-31
Not all `astropy.time` formats appear to be supported correctly.'


WilliamJamieson:
In looking at the coverage reports for `asdf-astropy` I noticed that the `astropy.time` time-formats covered by: https://github.com/astropy/asdf-astropy/blob/816dc57584a3735286e9d73f7d622570ecfc0507/asdf_astropy/converters/time/time.py#L22-L27 are not being properly tested by the `asdf-astropy` unit tests. However, when I added test-cases to cover these formats, I discovered that times using these formats do not round-trip through asdf using `asdf-astropy`. Note that this extends to when one uses `astropy.io.misc.asdf` as well.

We claim that `asdf-astropy` will produce an identical `astropy` object to that which was serialized for all supported `astropy` objects. This is clearly not happening at this time.

WilliamJamieson:
Tagging @mhvk and @taldcroft, for their advice on refactoring the time converter: https://github.com/astropy/asdf-astropy/blob/816dc57584a3735286e9d73f7d622570ecfc0507/asdf_astropy/converters/time/time.py#L14-L104 in general. Any advice on improving/simplifying this object would be great.

mhvk:
@WilliamJamieson - in newer astropy, the way to ensure precision is not lost when using a string representation is
```
from astropy.time import Time
t = Time(50000., 1/9, format='mjd')
t.to_value('mjd', 'str')
# '50000.1111111111111111'
```
Note that this will round-trip for MJD, but not for BYEAR, since the latter involves a calculation that uses a single `float`. But at least it should be considerably closer.

WilliamJamieson:
@mhvk, in an offline conversation with @taldcroft, it was suggested that maybe we should use something like `Time.info._represent_as_dict()`, which should be "stable". This would let us serialize the JD1/JD2 representation that astropy is using to represent the given time, which should round trip?

If this seems like an option, we might require something like a `to_dict` (or `to_yaml`) and `from_dict` method to be added to `astropy.time.Time` which we could then use a new `Converter` and `schema` for.

mhvk:
@WilliamJamieson - I think explicitly using the methods astropy uses internally to serialize objects makes sense. I'm not quite sure that this would need a method on `Time` itself, especially since you may well want a similar procedure for `SkyCoord`. Perhaps instead we should just make our current machinery public? (Could be functions or public methods on `.info`.)

WilliamJamieson:
[...]

PR #86, opened 2022-05-31
Update the time converter and its tests

WilliamJamieson:
In issue #85, I found several concerning things about how `asdf-astropy` handles converting and reading `time` objects. This PR attempts to resolve those.


mhvk:
I think for some formats, like `BYEAR`, it is not possible to round-trip currently if one insists on storing `.byear` itself, but for others like MJD one can use `t.to_value(format, 'str')` and one should be able to round-trip. For all formats, that incantation should provide the best possible estimate.

Note that the present tests here are not very thorough: many of the times chosen are exactly representable, so round-trip easily even if there is a general problem. In `astropy.time`, we run more thorough round-trip tests using `hypothesis` - see https://github.com/astropy/astropy/blob/main/astropy/time/tests/test_precision.py - might it be an idea to copy some of that framework over?

WilliamJamieson:
@mhvk, thanks for the suggestion. I think that adding hypothesis testing for `astropy.time` and possibly other parts of `astropy` would make an excellent future improvement to the testing of `asdf-astropy`.

For now however, I think this change is sufficient to cover the current issues discovered with the `astropy.time` serialization.

WilliamJamieson:
> LGTM, though I'm not familiar with all the intricacies of the Time class. I assume there is an update to the corresponding schema for the time tag.

There will be shortly.

PR #97, opened 2022-06-30
TST: pre-commit bot

WilliamJamieson:
This should break pre-commit-bot and not produce an auto-commit

Demo for astropy/astropy#13320

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,13 +7,13 @@ repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.1.0
   hooks:
-  - id: check-added-large-files
-  - id: check-case-conflict
-  - id: check-yaml
-    args: ["--unsafe"]
-  - id: debug-statements
-  - id: end-of-file-fixer
-  - id: trailing-whitespace
+    - id: check-added-large-files
+    - id: check-case-conflict
+    - id: check-yaml
+      args: ["--unsafe"]
+    - id: debug-statements
+    - id: end-of-file-fixer
+    - id: trailing-whitespace
 
 - repo: https://github.com/asottile/pyupgrade
   rev: v2.34.0
--- a/asdf_astropy/extensions.py
+++ b/asdf_astropy/extensions.py
@@ -30,8 +30,7 @@ TRANSFORM_CONVERTERS = [
     CompoundConverter(),
     # astropy.modeling.functional_models
     SimpleTransformConverter(
-        ["tag:stsci.edu:asdf/transform/airy_disk2d-*"],
-        "astropy.modeling.functional_models.AiryDisk2D",
+        ["tag:stsci.edu:asdf/transform/airy_disk2d-*"], "astropy.modeling.functional_models.AiryDisk2D",
     ),
     SimpleTransformConverter(
         ["tag:stsci.edu:asdf/transform/box1d-*"],
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,8 @@ write_to = "asdf_astropy/_version.py"
 
 [tool.black]
 line-length = 120
+check = true
+diff = true
 force-exclude = '''
 ^/(
   (


pllim:
I wish the failure would be more informative.

```
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted asdf_astropy/extensions.py

All done! ✨ 🍰 ✨
1 file reformatted, 66 files left unchanged.
```

Meanwhile, in the code style job that it is supposed to replace:

```
pre-commit hook(s) made changes.
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/asdf_astropy/extensions.py b/asdf_astropy/extensions.py
index 43e7e[57](https://github.com/astropy/asdf-astropy/runs/7138128092?check_suite_focus=true#step:5:58)..69d7380 100644
--- a/asdf_astropy/extensions.py
+++ b/asdf_astropy/extensions.py
@@ -30,7 +30,8 @@TRANSFORM_CONVERTERS = [
     CompoundConverter(),
     # astropy.modeling.functional_models
     SimpleTransformConverter(
-        ["tag:stsci.edu:asdf/transform/airy_disk2d-*"], "astropy.modeling.functional_models.AiryDisk2D",
+        ["tag:stsci.edu:asdf/transform/airy_disk2d-*"],
+        "astropy.modeling.functional_models.AiryDisk2D",
     ),
     SimpleTransformConverter(
         ["tag:stsci.edu:asdf/transform/box1d-*"],
```

nstarman:
@WilliamJamieson  Can the `log_file` + `verbose` be used to get a better error trace?

WilliamJamieson:
> @WilliamJamieson Can the `log_file` + `verbose` be used to get a better error trace?
> 
> https://pre-commit.com/#config-log_file

I got it to sort of work as in this
Screen Shot 2022-07-01 at 5 14 29 PM,

however it required setting the `diff` and `check` values for the black config, which turn off auto-fixing locally and probably via the update method from pre-commit. Therefore, it is pretty useless...

Please feel free to push your own commits onto this branch to see what works.

WilliamJamieson:
Also, the above is only tested for black output. This will have to be tested and worked through for every check.

nstarman:
> However to accomplish this I had to set the diff and check values for the black config, which turn off auto-fixing locally and probably via the update method from pre-commit (yes I could pass them as arguments through pre-commit, but the auto-fixing feature will still be broken). Therefore, it is pretty useless...

I'm not sure I see what the problem is. In my mind we either want a full trace OR to auto-fix — we never really need both. 
If we have auto-fixing, why would we need to see the trace? The fix is in the git history as a commit.
If we don't have auto-fixing then we need to see the trace. In this case we definitionally don't care if the auto fixing is broken.
Am I getting the square right?

| | auto fix on | off |
| --- | --- | --- |
| see trace | don't care about trace because already fixed | SGTM — we need (and have) the trace to fix the problem |
| don't see | SGTM — don't care about trace because already fixed | This would be bad. But AFAIK we never run into this |

WilliamJamieson:
The issue is it breaks auto-fixing for local development.

nstarman:
That is inconvenient.

nstarman:
I've figured out a solution, but it's painful.
We need a fork of black and to change the ``id`` to e.g. ``black-check``. We have two jobs — ``black`` and ``black`` check and in pre-commit.ci skip the jobs that fix using ``skip: [black]``.
It works, but is less good than just running in tox. The best option would be if ``pre-commit.ci`` had an ``args`` option.

WilliamJamieson:
Your thought sounds reasonable, but why skip the `[black]` job at all? Even though we have [`autofix_prs`](https://pre-commit.ci/#configuration-autofix_prs) disabled, it still provides the ability to fix the PR (if possible) by commenting with `pre-commit.ci autofix`. Which to be useful, would require the job to be able to generate code changes.

Also, as noted here: https://github.com/pre-commit-ci/issues/issues/134#issuecomment-1173192587 the output is quite truncated, so having the pre-commit page attempt to describe errors is probably a loosing proposition. I'm not sure the pre-commit CI bot will necessarily replace the `tox` based `Code Style Checks` (in this case) job, but provide additional functionality, such as

1. The option of automatically fixing PRs for user, even if it does not automatically fix them (again the `pre-commit.ci autofix` comment creates an autofix commit on the PR).
2. The creation of automatic PRs updating the `.pre-commit-config.yaml` file.
3. A fast code style check of a PR independent of our other CI jobs.

I suggest the current code style check can be turned into a check which gets run when a label (such as `Check Code Style`) is applied to a PR (such as when the automated fixing fails).

pllim:
I am a little lost on where this is going. If we still agree on autofix PR being opt-in only, then the CI check should always run, not just with label. So effectively, this bot is just one more addition to the army of bots/checks we already have, not replacing anything.

WilliamJamieson:
> I am a little lost on where this is going. If we still agree on autofix PR being opt-in only, then the CI check should always run, not just with label. So effectively, this bot is just one more addition to the army of bots/checks we already have, not replacing anything.

First, if we use pre-commit.ci the CI will always run via the pre-commit.ci bot; where it will run the equivalent checks as the current CI job. However, as pointed out by https://github.com/pre-commit-ci/issues/issues/134#issuecomment-1173192587, the displayed output of the bot's job is quickly truncated meaning all the above effort about improving its output is of limited utility. My suggestion is simply to preserve the existing code style job by modifying it to simply be activated via a PR label as necessary. This way if users don't want to opt into allowing the bot to apply a fix or the box cannot fix the code, there is a mechanism to tell users what is wrong in detail which is not truncated (or alternatively we just tell them to run it locally and fix the results). Ultimately, we should preserve the ability of both the bot (via opt-in) and pre-commit in general (for local use as a git hook) to automatically generate and apply changes to the code. 

Note that I personally think we should leave the pre-commit.ci bot in its default configuration and let it automatically fix the PRs. The argument that it is hostile to new users to have it automatically change their code is a bit hollow to me. Because then their PR is failing for some reason and they have either make some seemingly arbitrary but precise code changes by hand or learn how to setup and run all these tools themselves. In my opinion, it is better to just do the fix automatically rather than having contributors and maintainers dance around a bunch of options to fix the issue.

WilliamJamieson:
https://github.com/astropy/astropy/issues/13320 is now complete closing this.

PR #110, opened 2022-08-29
Increase test coverage of transforms

WilliamJamieson:
Increase the test coverage of transforms

### DIFF
--- a/asdf_astropy/converters/transform/compound.py
+++ b/asdf_astropy/converters/transform/compound.py
@@ -72,13 +72,13 @@ class CompoundConverter(TransformConverterBase):
 
         left = node["forward"][0]
         if not isinstance(left, Model):
-            raise TypeError("Unknown left model type '{}'".format(node["forward"][0]._tag))
+            raise TypeError(f"Unknown left model type '{node['forward'][0]._tag}'")
 
         right = node["forward"][1]
         if (oper == "fix_inputs" and not isinstance(right, dict)) or (
             oper != "fix_inputs" and not isinstance(right, Model)
         ):
-            raise TypeError("Unknown right model type '{}'".format(node["forward"][1]._tag))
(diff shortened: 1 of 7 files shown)


asdf_astropy/converters/transform/compound.py
```diff
@@ -72,11 +72,13 @@ def from_yaml_tree_transform(self, node, tag, ctx):
 
         left = node["forward"][0]
         if not isinstance(left, Model):
-            raise TypeError("Unknown model type '{}'".format(node["forward"][0]._tag))
+            raise TypeError("Unknown left model type '{}'".format(node["forward"][0]._tag))
 
         right = node["forward"][1]
-        if not isinstance(right, Model) and not (oper == "fix_inputs" and isinstance(right, dict)):
-            raise TypeError("Unknown model type '{}'".format(node["forward"][1]._tag))
+        if (oper == "fix_inputs" and not isinstance(right, dict)) or (
```
perrygreenfield, line 78: Could you not combine these two into one by using `isinstance(right, (dict, Model))`?

perrygreenfield (commented):
one minor comment

asdf_astropy/converters/transform/compound.py
WilliamJamieson: I don't think so, the issue is if its `fix_inputs` then it must be a dictionary. If it is not `fix_inputs` then it must be a model. You cannot have a `fix_inputs` with a model. That is why there are two statements with an `or` between them.

WilliamJamieson: I'm not sure if this case would pass the validation for an asdf file, but the original error was attempting to protect against these cases except you could pass a `fix_inputs` with two models. Which would cause it to fail in a very strange way. This covers that case now.

perrygreenfield (approved):
LGTM

PR #116, opened 2022-09-20
Switch to open-astronomy workflow

WilliamJamieson:
Switch to running CI with open-astronomy workflow.


perrygreenfield (approved):
LGTM, though I am not qualified yet to check ci.yml (so someone else has to do that)

PR #125, opened 2022-10-20
Bugfix for `memmap` not working with `Quantity`

WilliamJamieson:
Based on the discussion in astropy/astropy#13864, I found that we need to add the `copy=False` flag to the array constructor and arrange for a more complicated test for `memmap` than we normally require.

resolves astropy/astropy#13864

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@
 
 - Update citations. [#111]
 - Switch to using ``pyproject.toml`` for package configuration. [#106]
+- Fix bug with ``memmap`` of ``Quantity`` objects. [#125]
 
 0.2.2 (2022-08-22)
 ------------------
--- a/asdf_astropy/converters/unit/quantity.py
+++ b/asdf_astropy/converters/unit/quantity.py
@@ -27,4 +27,4 @@ class QuantityConverter(Converter):
             # and documentation.
             value = value._make_array()
 
-        return Quantity(value, unit=node["unit"])
+        return Quantity(value, unit=node["unit"], copy=False)
--- a/asdf_astropy/converters/unit/tests/test_quantity.py
+++ b/asdf_astropy/converters/unit/tests/test_quantity.py
@@ -65,3 +65,37 @@ quantity: !unit/quantity-1.1.0
     with asdf.open(buff) as af:
         assert_array_equal(af["quantity"].value, np.array([1.0, 2.0, 3.0, 4.0]))
         assert af["quantity"].unit.is_equivalent(units.km)
+
+
+def test_memmap(tmp_path):
+    """
+    Test that memmap works with quantities. Unfortunately, this is not a simple `isinstance(obj, np.memmap)`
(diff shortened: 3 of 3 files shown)


WilliamJamieson:
@mhvk Thank you for your note in https://github.com/astropy/astropy/issues/13864#issuecomment-1286009884.

Would you mind reviewing these changes?

CHANGES.rst
```diff
@@ -3,6 +3,7 @@
 
 - Update citations. [#111]
 - Switch to using ``pyproject.toml`` for package configuration. [#106]
+- Fix bug with ``memmap`` of ``Qauntity`` objects. [#125]
```
pllim, line 6:
```suggestion
- Fix bug with ``memmap`` of ``Quantity`` objects. [#125]
```

asdf_astropy/converters/unit/quantity.py
```diff
@@ -27,4 +27,4 @@ def from_yaml_tree(self, node, tag, ctx):
             # and documentation.
             value = value._make_array()
 
-        return Quantity(value, unit=node["unit"])
+        return Quantity(value, unit=node["unit"], copy=False)
```
pllim, line 30: Do we also need `subok`? Maybe @mhvk can confirm.

```suggestion
        return Quantity(value, unit=node["unit"], copy=False, subok=True)
```

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -65,3 +65,72 @@ def test_read_array_value():
     with asdf.open(buff) as af:
         assert_array_equal(af["quantity"].value, np.array([1.0, 2.0, 3.0, 4.0]))
         assert af["quantity"].unit.is_equivalent(units.km)
+
+
+def test_memmap(tmp_path):
+    """
+    Test that memmap (copy_arrays=False) works with quantities.
+
+    Unfortunately, this is not a simple `isinstance(obj, np.memmap)`
```
mhvk, line 74: In a way I guess it is OK to test just on the actual effect -- certainly, for `np.memmap`, something can be an instance without an actual file backing it.

mhvk:
@WilliamJamieson - the change looks logical and the tests good!

asdf_astropy/converters/unit/quantity.py
WilliamJamieson: Correct, this will only be handling the two types listed here: https://github.com/WilliamJamieson/asdf-astropy/blob/3888a86746953afee37b226c73ff96f22e7e2dea/asdf_astropy/converters/unit/quantity.py#L9-L12

nden: Soon we'll need to serialize `coordinates.SpectralQuantity` but I guess all we need to do then is add it to the above list.

nden (approved):
LGTM

perrygreenfield (approved):
LGTM

PR #131, opened 2022-11-18
Bugfix for `Quantity.dtype`

### LINKED ISSUE #130: `dtype` is not preserved roundtrip by `Quantity`
We currently do not preserve the `dtype` for `Quantity` objects roundtrip. E.G.

```python
import asdf
import astropy.units as u

amp33 = u.Quantity(np.zeros((5, 5), dtype=np.uint16), dtype=np.uint16)
print(f"{amp33.dtype=}")

af = asdf.AsdfFile({"amp33": amp33})
af.write_to('test.asdf')

with asdf.open('test.asdf', mode="rw") as ff:
    new_amp33 = ff["amp33"]

print(f"{new_amp33.dtype=}")
```
returns
```
amp33.dtype=dtype('uint16')
new_amp33.dtype=dtype('float64')
```

So the `dtype` is not making it round trip. I believe this is because `Quantity` automatically casts the `dtype` to `float64` unless otherwise specified. This means the `QuantityConverter` needs to add a `dtype` argument here: https://github.com/astropy/asdf-astropy/blob/9b5bd0ff2a0fd2fc9b90ccab60e129840747c8aa/asdf_astropy/converters/unit/quantity.py#L30 matching the `dtype` of `value` if it is a `ndarray`.

WilliamJamieson:
Bugfix so that the `dtype` of a `Quantity` object is preserved.

Fixes #130

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@
 - Fix bug with ``memmap`` of ``Quantity`` objects. [#125]
 - Drop support for ``numpy-1.18``. [#116]
 - Fix bug with ``str`` representations of ``astropy.time`` objects. [#132]
+- Fix bug in preserving the ``dtype`` of ``Quantity`` objects. [#131]
 
 0.2.2 (2022-08-22)
 ------------------
--- a/asdf_astropy/converters/unit/quantity.py
+++ b/asdf_astropy/converters/unit/quantity.py
@@ -13,20 +13,25 @@ class QuantityConverter(Converter):
     ]
 
     def to_yaml_tree(self, obj, tag, ctx):
-        return {
+        node = {
             "value": obj.value,
             "unit": obj.unit,
         }
+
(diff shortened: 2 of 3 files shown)


nden:
The error in the failing test is real. Do we need to support numpy 1.18?

WilliamJamieson:
> The error in the failing test is real. Do we need to support numpy 1.18?

No we should have dropped support for it in July.

WilliamJamieson:
> The error in the failing test is real. Do we need to support numpy 1.18?

I think this is better fixed in #116, after which we can rebase and drop the changes in 78e50b1.

perrygreenfield (approved):
Does the schema for quantities need to be updated to handle dtype or is that already there?

WilliamJamieson:
> Does the schema for quantities need to be updated to handle dtype or is that already there?

It was added by asdf-format/asdf-standard#351.

PR #132, opened 2022-11-21
Bugfix for time strings

### LINKED ISSUE #128: different output compared to astropy.io.misc.asdf
As discussed in this PR https://github.com/spacetelescope/stdatamodels/pull/104 astropy.io.misc.asdf and asdf-astropy will output different yaml for certain time formats.

It appears that asdf-astropy will always output an object even when a string is read (as seen in this gist: https://gist.github.com/braingram/b369a6326e299088fb98bcef38a01c7d

I don't necessarily see this as an issue as the object format is more descriptive and less ambiguous than the string formatted time (at the expense of slightly more yaml). It may however create issues with downstream packages.

If this is intended behavior, is this documented?

WilliamJamieson:
This PR returns some pure string outputs from the time converter under some circumstances as requested in #128 reverting some of the changes in #86. With this PR:

- `iso`
- `jyear_str`
- `byear_str`
- `yday`

will be written purely as strings. Before #86 the:

- `isot`
- `jyear`
- `byear`

formats would also be written as pure strings; however, these will no round-trip the format as they would be deserialize as `iso`, `jyear_str`, and `byear_str` formats respectively.

Fixes #128

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@
 - Switch to using ``pyproject.toml`` for package configuration. [#106]
 - Fix bug with ``memmap`` of ``Quantity`` objects. [#125]
 - Drop support for ``numpy-1.18``. [#116]
+- Fix bug with ``str`` representations of ``astropy.time`` objects. [#132]
 
 0.2.2 (2022-08-22)
 ------------------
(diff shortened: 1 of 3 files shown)


WilliamJamieson:
Note this should partially resolve what is being fixed by `spacetelescope/stdatamodels#104`, but not completely. Instead of removing the `astropy.time.Time` object from being used, it simply needs to be forced to format as `iso`.

perrygreenfield (approved):
LGTM

PR #139, opened 2022-11-30
Turn Warnings into errors

WilliamJamieson:
Turns warnings into errors for better detection of emerging issues.

### DIFF
--- a/asdf_astropy/converters/time/tests/test_time.py
+++ b/asdf_astropy/converters/time/tests/test_time.py
@@ -47,6 +47,7 @@ def create_times():
 
 @pytest.mark.parametrize("time", create_times())
 @pytest.mark.parametrize("version", asdf.versioning.supported_versions)
+@pytest.mark.filterwarnings("ignore:ERFA function.*")  # ERFA warning emitted for plot_date format
 def test_serialization(time, version, tmp_path):
     file_path = tmp_path / "test.asdf"
     with asdf.AsdfFile(version=version) as af:
--- a/asdf_astropy/converters/time/tests/test_time.py
+++ b/asdf_astropy/converters/time/tests/test_time.py
@@ -34,7 +34,7 @@ def create_times():
             format="jyear_str",
         ),
         Time(51544.0, format="mjd"),
-        Time(730120.0003703703, format="plot_date"),
+        Time(Time("2000-01-01T00:00:00.000").plot_date, format="plot_date"),  # astropy/astropy#14080
         Time(np.arange(100), format="unix"),
         Time(946684800.0, format="unix_tai"),
         Time("2000:001:00:00:00.000", format="yday"),
@@ -47,7 +47,6 @@ def create_times():
 
 @pytest.mark.parametrize("time", create_times())
 @pytest.mark.parametrize("version", asdf.versioning.supported_versions)
-@pytest.mark.filterwarnings("ignore:ERFA function.*")  # ERFA warning emitted for plot_date format
 def test_serialization(time, version, tmp_path):
     file_path = tmp_path / "test.asdf"
     with asdf.AsdfFile(version=version) as af:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,19 +77,6 @@ jobs:
         - linux: py39-test-numpy121
         - linux: py39-test-numpy122
 
-  warnings:
-    needs: [core, asdf-transform-schemas]
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
-    with:
-      cache-path: ~/.cache/pip
-      cache-key: pip-${{ needs.setup.outputs.requirements-hash }}
-      cache-restore-keys: |
-        pip-
-      # Any env name which does not start with `pyXY` will use this Python version.
-      default_python: '3.9'
-      envs: |
-        - linux: warnings
-
   wheel_building:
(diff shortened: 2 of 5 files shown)


asdf_astropy/converters/time/tests/test_time.py
```diff
@@ -47,6 +47,7 @@ def create_times():
 
 @pytest.mark.parametrize("time", create_times())
 @pytest.mark.parametrize("version", asdf.versioning.supported_versions)
+@pytest.mark.filterwarnings("ignore:ERFA function.*")  # ERFA warning emitted for plot_date format
```
WilliamJamieson, line 50: @mhvk and @taldcroft I had to add this filter in because `Time(730120.0003703703, format="plot_date")` (line 37 above) was causing this warning
```
erfa.core.ErfaWarning: ERFA function "d2dtf" yielded 1 of "dubious year (Note 5)
```
to be emitted here: https://github.com/astropy/asdf-astropy/blob/cf297c37db6e14ecb0d0b33ae0d47b5ba20766b8/asdf_astropy/converters/time/time.py#L37

when serializing. Is this an issue we should be concerned about or is ignoring this warning okay?

mhvk: I think it is fair to ignore the warning, though my sense would be to use a less dubious plot date (something around year 2000 or so) so that one doesn't get it in the first place.

WilliamJamieson: Unfortunately, I tried this see c78e4ef,  which produces the warning I mentioned: https://github.com/astropy/asdf-astropy/actions/runs/3586673705/jobs/6036160520.

WilliamJamieson: Note the `10957.0` was generated for `Time(10957.0, format="plot_date")`, by converting `Time("2000-01-01T00:00:00.000")` to the `plot_date` format.

mhvk: Something is fishy here:
```
In [3]: Time(10957.0, format="plot_date").isot
/data/mhvk/venv/numpy-dev/lib/python3.10/site-packages/erfa/core.py:154: ErfaWarning: ERFA function "d2dtf" yielded 1 of "dubious year (Note 5)"
  warnings.warn('ERFA function "{}" yielded {}'.format(func_name, wmsg),
Out[3]: '30-12-31T00:00:00.000'
```
And clearly what you had originally looks good:
```
In [5]: Time("2000-01-01T00:00:00.000").plot_date
Out[5]: 730120.0
```

Indeed, I get no warnings with the number you have above,
```
In [6]: Time(730120.0003703703, format='plot_date').isot
Out[6]: '2000-01-01T00:00:32.000'
```

I'm very confused!!

WilliamJamieson: I am very confused too... When I run
```
In [1]: from astropy.time import Time
In [2]: Time("2000-01-01T00:00:00.000").plot_date
Out[2]: 10957.0
```

mhvk: Huh??? I get
```
In [1]: %astropy
Out[1]: '5.1.1'

In [2]: Time("2000-01-01T00:00:00.000").plot_date
Out[2]: 730120.0
```

But, super weirdly, this is only under numpy's development version...  
```
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.25.0.dev0+61.g3c2717c9a'

In [3]: import astropy

In [4]: astropy.__version__
Out[4]: '5.1.1'

In [5]: from astropy.time import Time

In [6]: Time("2000-01-01T00:00:00.000").plot_date
Out[6]: 730120.0
```

When I try it with the regular Debian-provided numpy, it works fine
```
In [3]: import numpy as np

In [4]: np.__version__
Out[4]: '1.21.5'

In [5]: import astropy

In [6]: astropy.__version__
Out[6]: '5.1.1'

In [7]: from astropy.time import Time

In [8]: Time("2000-01-01T00:00:00.000").plot_date
Out[8]: 10957.0
```

mhvk: To add to the confusion, according to the docstring, "730120.0003703703 is midnight on January 1, 2000." -- see https://github.com/astropy/astropy/blob/27cf81dff23876fa871786d7ee710cb4c120c413/astropy/time/formats.py#L864-L879

But looking at the code there, I suspect the problem is that the zero point changed in matplotlib, so this may be matplotlib-version dependent... Indeed, if I install matplotlib, then the answer is 10957.0

So, bottom line may be that the value one should use depends on whether or not matplotlib is installed -- which is pretty terrible!

mhvk: I think we should address this in astropy; see https://github.com/astropy/astropy/issues/14080

For the issue here, could it be that tests are run both with and without matplotlib, so each version fails on one of the two (either year way before it makes sense, or way beyond). Possibly, the best route is to use as input
```
Time(Time("2000-01-01T00:00:00.000").plot_date, format='plot_date')
```
(with a comment of course!!)

WilliamJamieson: This is what I ended up doing. Hopefully it won't matter if I move astropy times to a custom schema/converter.

mhvk: Sounds good!

perrygreenfield (approved):
LGTM

PR #142, opened 2022-12-08
Add support for non-VOunits

### LINKED ISSUE #144: Support non-VOunits
(note that this issue has been opened after the PR to fix it)

@PaulHuwe has requested that we consider supporting non-VOunits which are part of astropy as a more general thing. For example I had to do a hack job in spacetelescope/roman_datamodels#109 in order to support a few non-VOunits for the Roman Space Telescope datamodels.

Initially, I was opposed to adding general support for non-VOunits to ASDF due to the possibility of confusion of unit symbol meanings. However, after some consideration I have come to the conclusion that we can indeed support non-VOunits which originate from astropy in a way that ensures that no confusion can occur.

My plan for doing this is simple, units will be written under two different unit tags:

- `tag:tag:stsci.edu:asdf/unit/unit-1.0.0` (the original tag) for VOunits.
- `tag:astropy.org:astropy/units/unit-1.0.0` for non-VOunits specified by astropy.

Support for this will be slightly complicated as the object types in astropy for units are not divided between VOunits and non-VOunits. However, theoretically a `select_tag` method can be implemented to distinguish them.

WilliamJamieson:
Adds support for non-VOunit serialization to a separate tag/schema than VOunits.

Fixes #144

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ------------------
 
 - Update pins for ``asdf``, ``asdf-coordinates-schemas``, ``numpy``, and ``packaging``. [#164]
+- Add serialization support for non-VOunits. [#142]
 
 0.3.0 (2022-11-29)
 ------------------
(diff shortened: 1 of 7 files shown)


perrygreenfield (approved):
LGTM

PR #152, opened 2022-12-15
Move downstream testing to using tox and openastronomy CI 

WilliamJamieson:


nden (approved):
I'm not qualified to review in detail. It looks good as far as I can tell.

PR #179, opened 2023-05-02
Remove astropy testing now that io.misc.asdf is gone.

WilliamJamieson:


braingram (approved):
Thanks for making these changes!

PR #180, opened 2023-05-02
Drop Python 3.8 support

WilliamJamieson:
Astropy has dropped python 3.8, asdf-astropy needs to follow suite.


braingram (changes requested):
Looks good but needs a changelog entry.

PR #181, opened 2023-05-23
Update `RepresentationConverter` for new class paths in astropy dev

braingram:
Astropy recently moved a few classes:

https://github.com/astropy/astropy/pull/14792
moved several classes previously in `astropy.coordinates.representation` to submodules. This broke the `RepresentationConverter`. This PR adds the new class paths (and retains the old ones for backwards compatibility).

~https://github.com/astropy/astropy/pull/14763
moved several `Geodetic` classes into `astropy.coordinates.representation`. The `__all__` of this module is used within the asdf-astropy tests to collect classes to test the `RepresentationConverter`. It does not appear that these `Geodetic` classes are currently supported by asdf-astropy so the test was updated to ignore these classes from the list of classes to test.~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ------------------
 
 - Drop support for Python 3.8 in accordance with NEP 29. [#180]
+- Update ``RepresentationConverter`` for new class paths in astropy [#181]
 
 0.4.0 (2023-03-20)
 ------------------
--- a/asdf_astropy/converters/coordinates/representation.py
+++ b/asdf_astropy/converters/coordinates/representation.py
@@ -19,6 +19,21 @@ class RepresentationConverter(Converter):
         "astropy.coordinates.representation.UnitSphericalCosLatDifferential",
         "astropy.coordinates.representation.UnitSphericalDifferential",
         "astropy.coordinates.representation.UnitSphericalRepresentation",
+        # classes were moved in https://github.com/astropy/astropy/pull/14792
+        "astropy.coordinates.representation.cartesian.CartesianDifferential",
+        "astropy.coordinates.representation.cartesian.CartesianRepresentation",
+        "astropy.coordinates.representation.cylindrical.CylindricalDifferential",
+        "astropy.coordinates.representation.cylindrical.CylindricalRepresentation",
+        "astropy.coordinates.representation.spherical.PhysicsSphericalDifferential",
+        "astropy.coordinates.representation.spherical.PhysicsSphericalRepresentation",
+        "astropy.coordinates.representation.spherical.RadialDifferential",
+        "astropy.coordinates.representation.spherical.RadialRepresentation",
+        "astropy.coordinates.representation.spherical.SphericalCosLatDifferential",
+        "astropy.coordinates.representation.spherical.SphericalDifferential",
+        "astropy.coordinates.representation.spherical.SphericalRepresentation",
+        "astropy.coordinates.representation.spherical.UnitSphericalRepresentation",
+        "astropy.coordinates.representation.spherical.UnitSphericalDifferential",
+        "astropy.coordinates.representation.spherical.UnitSphericalCosLatDifferential",
     ]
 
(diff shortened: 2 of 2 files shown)


asdf_astropy/converters/coordinates/tests/test_representation.py
```diff
@@ -6,8 +6,13 @@
 
 from asdf_astropy.testing.helpers import assert_representation_equal
 
+# https://github.com/astropy/astropy/pull/14763 moved Geodetic representations from
+# astropy.coordinates.earth to astropy.representations.geodetic we do not have
```
WilliamJamieson, line 10:
```suggestion
# astropy.coordinates.earth to astropy.coordinates.representations.geodetic we do not have
```

asdf_astropy/converters/coordinates/tests/test_representation.py
```diff
@@ -6,8 +6,13 @@
...
+# converters for these types so we skip the geodetic classes in these tests
```
WilliamJamieson, line 11: The `astropy.coordinates.representations.geodetic` objects seem to be subclasses of the same `BaseRepresentation` object, so they should already be serializable using the existing converter. Please explain why we need to ignore these? The only issue I see is that `asdf-coordinates-schemas` does not list the new geodetic types.

If this is necessary, please either add converter support for these objects or open an explicit issue to track the creation of converters for these objects.

WilliamJamieson (changes requested):

asdf_astropy/converters/coordinates/tests/test_representation.py
braingram: I opened an issue for this.

https://github.com/astropy/asdf-astropy/issues/182

I don't see any converter that lists `BaseRepresentation`. Which converter should be supporting these new classes?

WilliamJamieson: No, those are just grouping classes only, so that one can do `isinstance` checks.

braingram: As you pointed out in https://github.com/asdf-format/asdf-coordinates-schemas/pull/40#pullrequestreview-1440302867
additional updates will be required to add support for the Geodetic representations. I removed the skip for the tests against these classes from this PR.

This PR should now address class path changes introduced in https://github.com/astropy/astropy/pull/14792

pyproject.toml
```diff
@@ -18,7 +18,7 @@ dynamic = [
 ]
 dependencies = [
   "asdf>=2.13",
-  "asdf-coordinates-schemas>=0.1",
+  "asdf-coordinates-schemas@ git+https://github.com/WilliamJamieson/asdf-coordinates-schemas.git@add_geodetic", # "asdf-coordinates-schemas>=0.1",
```
WilliamJamieson, line 21: This should wait for `asdf-coordinates-schemas==0.3.0` to be released. It will have the necessary schema changes.

WilliamJamieson (changes requested):
This should be on hold until an `asdf-coordinates-schemas` release supporting this is made.

braingram:
@WilliamJamieson Can we split this into 2 PRs?

1. addressing the move of already supported classes to `astropy.coordinates.representation` submodules
2. adding support for `Geodetic` coordinate representations (as you noted this will have to wait for updates to `asdf-coordinates-schemas`)

WilliamJamieson:
> @WilliamJamieson Can we split this into 2 PRs?
> 
> 1. addressing the move of already supported classes to `astropy.coordinates.representation` submodules
> 2. adding support for `Geodetic` coordinate representations (as you noted this will have to wait for updates to `asdf-coordinates-schemas`)

You are welcome to do whatever you want.

nden (approved):
This looks good to me. The Geodetic classes are not released yet as far as I can tell. They can be added later. @WilliamJamieson Do you have any other concerns about this PR?

WilliamJamieson (approved):

PR #187, opened 2023-07-13
Make class variables immutable

WilliamJamieson:
An issue pointed out in #183 is that we have many of the classes which define mutable class variables. In general, mutable class variables should not be used because the state of that variable is shared across all instances. Thus they should be made to be immutable objects.

This PR corrects this issue for this package.

Note that this issue occurs mostly for the  `tags` and `types` attributes of the converters. These were turned into properties instead of tuples because ASDF does not support these values being anything other than lists (ASDF will literally segfault).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 - Drop support for Python 3.8 in accordance with NEP 29. [#180]
 - Update ``RepresentationConverter`` for new class paths in astropy [#181]
 
+- Fix mutable class variables being used by converters. [#187]
+
 0.4.0 (2023-03-20)
 ------------------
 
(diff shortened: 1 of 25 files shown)


braingram:
@WilliamJamieson Would you help me reproduce the segfault? Testing locally with several asdf version and extensions within asdf, custom extensions and ones in asdf-astropy I've so far failed to produce a segfault by changing the types and tags to tuples.

WilliamJamieson:
> @WilliamJamieson Would you help me reproduce the segfault? Testing locally with several asdf version and extensions within asdf, custom extensions and ones in asdf-astropy I've so far failed to produce a segfault by changing the types and tags to tuples.

You are correct, it was an independent bug I introduced.

PR #188, opened 2023-07-13
Attempt 2 Make class variables immutable

WilliamJamieson:
Attempt 2 at #187

### DIFF
--- a/asdf_astropy/converters/unit/tests/test_magunit.py
+++ b/asdf_astropy/converters/unit/tests/test_magunit.py
@@ -34,7 +34,7 @@ def create_magunits():
             else:
                 magunits.append(magunit)
 
-    return set(magunits)
+    return frozenset(magunits)
 
 
 @pytest.mark.parametrize("unit", create_magunits())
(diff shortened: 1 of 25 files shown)


asdf_astropy/converters/unit/tests/test_magunit.py
```diff
@@ -34,7 +34,7 @@ def create_magunits():
             else:
                 magunits.append(magunit)
 
-    return magunits
+    return set(magunits)
```
perrygreenfield, line 37: For immutability shouldn't `frozenset()` be used?

perrygreenfield (commented):
Other than one comment on sets, LGTM. But if this should be the general approach now, we should change the converter examples too in readthedocs.

asdf_astropy/converters/unit/tests/test_magunit.py
WilliamJamieson, line 37:
```suggestion
    return frozenset(magunits)
```

perrygreenfield (approved):
LGTM

PR #193, opened 2023-08-18
Drop oldest supported numpy from build-system

WilliamJamieson:
This is superfluous and was never needed.
It appears to have come from copy/paste from the old astropy package template


pllim (approved):
LGTM. Thanks for the quick fix!

PR #199, opened 2023-08-31
generate a unique uri for extensions that convert core types

braingram:
asdf-astropy currently implements converters for a number of core asdf types/tags. These converters are registered with an extension that is created in part using the core manifests provided by asdf-standard:
https://github.com/astropy/asdf-astropy/blob/38f76e19b85bf44b7d83d8f76f6705121e1f49ad/asdf_astropy/extensions.py#L520-L539

Because these extensions use the extension uri defined in the core manifests yet do not fully support all tags, this creates an issue where files can be written with incomplete extension history.

For example, using current development version of asdf, writing an empty tree produces the following file:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.0.0.dev591+g3c724f15}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf, version: 3.0.0.dev591+g3c724f15}
...
```
note the `asdf` software marked in the list of extensions in the history metadata.

Writing a file with an astropy Unit results in the following file:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.0.0.dev591+g3c724f15}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf_astropy._manifest.CompoundManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf-astropy, version: 0.4.1.dev46+g38f76e1.d20230831}
u: !unit/unit-1.0.0 m
...
```
note that the `asdf` extension is no longer listed.

This PR modifies the extension uri defined for asdf-astropy core extensions to be unique (replacing `asdf-format` in the uri with `astropy`), allowing asdf to include both softwares in the file history. The single unit file above generated with the changes included in this PR is as follows:
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.0.0.dev612+ge428962d}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf, version: 3.0.0.dev612+ge428962d}
  - !core/extension_metadata-1.0.0
    extension_class: asdf_astropy._manifest.CompoundManifestExtension
    extension_uri: asdf://astropy.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf-astropy, version: 0.4.1.dev46+g38f76e1.d20230831}
u: !unit/unit-1.0.0 m
...
```

A test was added (which does something similar to the above examples) to check for future regressions of the issue fixed in this PR.

Adding the above mentioned test revealed an issue with `test_no_astropy_import`. This needed to be addressed as depending on the test order the newly added test would fail on`import astropy.units` due to the bug in `test_no_astropy_import`. The modifications to `sys.modules` in `test_no_astropy_import` were changed to fix the bug. See the commit message for more details: 88be68c9638fb00ead1ef3e4036969faf4c7b029

`py39-test-oldestdep-parallels-cov` was also added to tox.ini as it otherwise will not run with the current tox.

### DIFF
--- a/asdf_astropy/tests/test_integration.py
+++ b/asdf_astropy/tests/test_integration.py
@@ -3,6 +3,7 @@ import sys
 from pathlib import Path
 
 import asdf
+import pytest
 import yaml
 
 
@@ -59,15 +60,49 @@ _ASTROPY_MODULES = [
 ]
 
 
+@pytest.fixture()
+def _clean_astropy_imports():
+    """Temporally unload all astropy modules used by asdf-astropy"""
+
+    # If any astropy or asdf_astropy modules are already imported
+    # remove them from sys.modules so we can later in this test check
+    # if those modules were imported. We have to store these as
+    # many submodules will reference or use types defined in other
+    # submodules
+    previous_modules = {}
+    for name in sys.modules.copy():
+        if name.startswith("asdf_astropy") or any(name.startswith(m) for m in _ASTROPY_MODULES):
+            previous_modules[name] = sys.modules[name]
+            del sys.modules[name]
+
+    # Register a module finder that just raises an exception if
+    # one of the tracked astropy modules is imported
+    class _Finder:
+        def find_spec(self, modulename, path=None, target=None):
+            if any(modulename.startswith(m) for m in _ASTROPY_MODULES):
+                msg = f"attempt to import astropy submodule({modulename}) during integration"
+                raise Exception(msg)  # noqa: TRY002
+
+    sys.meta_path.insert(0, _Finder())
+
+    # Setup complete return the test
+    yield
+
+    # Restore the previously imported modules. This is necessary as other code
+    # may have already used the modules we removed from the cache (astropy.constants
+    # defines some quantities).
+    for name in previous_modules:
+        sys.modules[name] = previous_modules[name]
+    if isinstance(sys.meta_path[0], _Finder):
+        sys.meta_path.pop(0)
+
+
+@pytest.mark.usefixtures("_clean_astropy_imports")
 def test_no_astropy_import():
     """
     Confirm that none of the ASDF plugins import astropy modules
     at import time.
     """
-
-    keys = [k for k in sys.modules if k.startswith("asdf_astropy") or any(k.startswith(m) for m in _ASTROPY_MODULES)]
-    for key in keys:
-        del sys.modules[key]
 
     from asdf_astropy import integration
 
--- a/asdf_astropy/tests/test_integration.py
+++ b/asdf_astropy/tests/test_integration.py
@@ -1,3 +1,4 @@
+import io
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


asdf_astropy/tests/test_integration.py
```diff
@@ -64,13 +65,63 @@ def test_no_astropy_import():
...
+
+        assert not any(k for k in sys.modules if any(k.startswith(m) for m in _ASTROPY_MODULES))
+    finally:
+        # Restore the previously imported modules. This is necessary as other code
+        # may have already used the modules we removed from the cache (astropy.constants
+        # defines some quantities).
+        for name in previous_modules:
+            sys.modules[name] = previous_modules[name]
+        if isinstance(sys.meta_path[0], _Finder):
+            sys.meta_path.pop(0)
```
WilliamJamieson, line 103: Most of this should be placed into a fixture as it distracts from what the actual test is. Instead do something like:
```python
@pytest.fixture()
def clean_astropy_imports():
    """Temporally unload all astropy modules used by asdf-astropy"""

    # If any astropy or asdf_astropy modules are already imported
    # remove them from sys.modules so we can later in this test check
    # if those modules were imported. We have to store these as
    # many submodules will reference or use types defined in other
    # submodules
    previous_modules = {}
    for name in sys.modules.copy():
        if name.startswith("asdf_astropy") or any(name.startswith(m) for m in _ASTROPY_MODULES):
            previous_modules[name] = sys.modules[name]
            del sys.modules[name]

    # Register a module finder that just raises an exception if
    # one of the tracked astropy modules is imported
    class _Finder:
        def find_spec(self, modulename, path=None, target=None):
            if any(modulename.startswith(m) for m in _ASTROPY_MODULES):
                msg = f"attempt to import astropy submodule({modulename}) during integration"
                raise Exception(msg)  # noqa: TRY002

    sys.meta_path.insert(0, _Finder())

    # Setup complete return to the test
    yield

    # Restore the previously imported modules. This is necessary as other code
    # may have already used the modules we removed from the cache (astropy.constants
    # defines some quantities).
    for name in previous_modules:
        sys.modules[name] = previous_modules[name]
    if isinstance(sys.meta_path[0], _Finder):
        sys.meta_path.pop(0)


@pytest.mark.usefixtures("clean_astropy_imports")
def test_no_astropy_import():
    """
    Confirm that none of the ASDF plugins import astropy modules
    at import time.
    """
    from asdf_astropy import integration

    integration.get_resource_mappings()
    integration.get_extensions()

    assert not any(k for k in sys.modules if any(k.startswith(m) for m in _ASTROPY_MODULES))
```

braingram: Good idea! Updated in: 4b193cf7eeb70ffa27f27dfd9cd62004555352a7 (ruff wanted a leading `_` to the fixture because of PT004).

WilliamJamieson (approved):
Should I plan to make a bugfix release once this is merged?

PR #214, opened 2023-10-30
Ignore `GeneralSersic2D` model until astropy 6 comes out.

WilliamJamieson:
astropy/astropy#15545 added the `GeneralSersic2D` model. There currently is not a schema for it until asdf-format/asdf#109 is merged and released along side `astropy` 6.0. This PR ignores this model when testing until upstream support is provided.


braingram (approved):
Thanks!

I believe the linked PR should be: https://github.com/asdf-format/asdf-transform-schemas/pull/109 (adding it here to make the xref).

PR #219, opened 2024-02-07
Update for ASDF standard 1.6.0

braingram:
This PR updates this asdf extension for new versions of:
- `asdf-standard`
- `asdf-coordinates-schemas`
- `asdf-transform-schemas`

`CompoundManifestExtension` was removed (as it is no longer required)

and updates the CI to account for:
- `astropy` dev requiring python 3.10
- remove testing against a bunch of random old numpy versions
- separate astropy-dev and numpy-dev testing
- missing python 3.12 testing

~a pytest upper pin was added to deal with: https://github.com/scientific-python/pytest-doctestplus/issues/243
when merged an issue can be opened to track removal of this pin~ EDIT: a new version of pytest-doctestplus was released that fixed the issue


braingram:
The `MaskedArray` failure with numpy 2.0 is unrelated and addressed in: https://github.com/asdf-format/asdf/pull/1769

nden (approved):
LGTM

PR #224, opened 2024-04-04
Update `copy` usage in `Quantity` converter to deal with astropy 6.1 changes

braingram:
astropy 6.1 changed the behavior of `copy`:
https://github.com/astropy/astropy/pull/16198

This breaks the `Quantity` converter and was not caught due to the `devdeps` testing against astropy "nightlies" which haven't been updated.

This PR adds a version check against numpy since that is the core source of the error (it looks like astropy will just pass es through `copy` to `array` instead of selectively calling `asarray`).

This PR also updates the CI to:
- remove astropy and transform 'lts' testing since 'lts' no longer exists
- adds a `predeps` job that runs against pre-releases

The masked array failure in the `predeps` should be fixed by:
https://github.com/asdf-format/asdf/pull/1769
which has not yet been released.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,7 @@ jobs:
       envs: |
         - linux: py39-parallel-cov
         - linux: py311-test-devdeps-parallel-cov
-        - linux: py39-astropylts-parallel-cov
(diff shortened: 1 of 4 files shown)


nden (approved):
lgtm

PR #229, opened 2024-05-09
Strip `factor=None` in equivalency converter for spectral density

braingram:
Equivalencies are serialized by storing the `kwargs`:
https://github.com/astropy/asdf-astropy/blob/1b286108042c1201199e9931cd2a5b52cc57bdce/asdf_astropy/resources/schemas/units/equivalency-1.1.0.yaml#L19

The `factor` keyword argument to `spectral_density` is deprecated in astropy 7.0 which means:
- any file that contains a `spectral_density` that was originally created with a `factor` will produce a warning when loaded
- these files will be unreadable when `factor` is removed

This PR strip `factor` when it's `None` to avoid the deprecation warning for astropy 7 and only tests `factor` for astropy versions older than 7.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+0.6.2 (unreleased)
+------------------
+
+- strip None factor for spectral_density in equivalency converter
+  to avoid deprecation warnings for astropy 7. [#229]
+
+
 0.6.1 (2024-04-05)
 ------------------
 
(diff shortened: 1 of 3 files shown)


pllim:
Why ignore the deprecation instead of just stop using deprecated code?

braingram:
> Why ignore the deprecation instead of just stop using deprecated code?

This PR is still draft. We will likely ignore the warning in the test to continue to test that we can serialize an equivalency created with a "factor" (which will produce a deprecation warning that would lead to a test failure). We can drop the test and the ignore when "factor" is removed from astropy.

pllim:
I am very surprised that `factor` is used at all. It was discouraged a very long time ago in the API doc.

PR #230, opened 2024-07-18
replace usages of ``copy_arrays`` with ``memmap`` for ``asdf>=3.1.0``

zacharyburnett:
follow-on to https://github.com/asdf-format/asdf/pull/1797

### DIFF
--- a/asdf_astropy/converters/unit/tests/test_quantity.py
+++ b/asdf_astropy/converters/unit/tests/test_quantity.py
@@ -1,16 +1,15 @@
-import importlib.metadata
-
 import asdf
 import numpy as np
 import pytest
 from asdf.testing import helpers
 from astropy import units
 from astropy.units import Quantity
+from astropy.utils.introspection import minversion
 from numpy.testing import assert_array_equal
 
 
 def asdf_open_memory_mapping_kwarg(memmap: bool) -> dict:
-    if tuple(int(part) for part in importlib.metadata.version("asdf").split(".")) >= (3, 1, 0):
+    if minversion("asdf", "3.1.0"):
         return {"memmap": memmap}
     else :
         return {"copy_arrays": not memmap}
--- a/asdf_astropy/converters/unit/tests/test_quantity.py
+++ b/asdf_astropy/converters/unit/tests/test_quantity.py
@@ -1,3 +1,5 @@
+import importlib.metadata
+
 import asdf
 import numpy as np
 import pytest
@@ -5,6 +7,13 @@ from asdf.testing import helpers
 from astropy import units
 from astropy.units import Quantity
 from numpy.testing import assert_array_equal
+
+
+def asdf_open_memory_mapping_kwarg(memmap: bool) -> dict:
+    if tuple(int(part) for part in importlib.metadata.version("asdf").split(".")) >= (3, 1, 0):
+        return {"memmap": memmap}
+    else :
+        return {"copy_arrays": not memmap}
 
 
 def create_quantities():
@@ -77,7 +86,7 @@ quantity: !unit/quantity-1.1.0
 
 def test_memmap(tmp_path):
     """
-    Test that memmap (copy_arrays=False) works with quantities.
+    Test that memmap (memmap=True) works with quantities.
 
     Unfortunately, this is not a simple `isinstance(obj, np.memmap)`
     Instead it requires a more complicated check.
@@ -95,7 +104,7 @@ def test_memmap(tmp_path):
         af.write_to(file_path)
 
     # Update a value in the ASDF file
-    with asdf.open(file_path, mode="rw", copy_arrays=False) as af:
+    with asdf.open(file_path, mode="rw", **asdf_open_memory_mapping_kwarg(memmap=True)) as af:
         assert (af.tree["quantity"] == quantity).all()
         assert af.tree["quantity"][-1, -1] != new_value
 
@@ -105,7 +114,7 @@ def test_memmap(tmp_path):
         assert (af.tree["quantity"] != quantity).any()
         assert (af.tree["quantity"] == new_quantity).all()
 
-    with asdf.open(file_path, mode="rw", copy_arrays=False) as af:
+    with asdf.open(file_path, mode="rw", **asdf_open_memory_mapping_kwarg(memmap=True)) as af:
         assert af.tree["quantity"][-1, -1] == new_value
         assert (af.tree["quantity"] != quantity).any()
         assert (af.tree["quantity"] == new_quantity).all()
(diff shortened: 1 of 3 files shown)


braingram:
Hmmm... the oldestdeps should have failed here since `asdf==2.13` (our minimum) doesn't have memmap:
https://github.com/astropy/asdf-astropy/blob/dab5b4d73ce3f75e60e5b29e75eb3e811dfd8319/pyproject.toml#L21
It doesn't seem to even be running `minimum_dependencies` :-/ and is picking up asdf 3.3.0
https://github.com/astropy/asdf-astropy/actions/runs/9993708226/job/27621947749?pr=230#step:10:24

@zacharyburnett what do you think we should do here? The usage is only in tests so run-time asdf-astropy doesn't need asdf>=3.1.0 but if we fix our oldest deps it will start failing (since it uses the runtime requirements). The tests would also fail for anyone who packages this repository using the run-time requirements to run tests (like the asdf issue that came up yesterday). We could update the tests to check the asdf version (and use either `copy_arrays` or `memmap`) but I'm not sure if it's really worth it (although that would allow us to fix the oldestdeps and hopefully have it not fail).

zacharyburnett:
> Hmmm... the oldestdeps should have failed here since `asdf==2.13` (our minimum) doesn't have memmap:
> 
> https://github.com/astropy/asdf-astropy/blob/dab5b4d73ce3f75e60e5b29e75eb3e811dfd8319/pyproject.toml#L21
> 
> 
> It doesn't seem to even be running `minimum_dependencies` :-/ and is picking up asdf 3.3.0
> https://github.com/astropy/asdf-astropy/actions/runs/9993708226/job/27621947749?pr=230#step:10:24
> @zacharyburnett what do you think we should do here? The usage is only in tests so run-time asdf-astropy doesn't need asdf>=3.1.0 but if we fix our oldest deps it will start failing (since it uses the runtime requirements). The tests would also fail for anyone who packages this repository using the run-time requirements to run tests (like the asdf issue that came up yesterday). We could update the tests to check the asdf version (and use either `copy_arrays` or `memmap`) but I'm not sure if it's really worth it (although that would allow us to fix the oldestdeps and hopefully have it not fail).

I think the only two options we have that maintain green status are
- remove oldestdeps testing or
- use `copy_arrays` or `memmap` based on version

I'm inclined toward the latter.

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -7,6 +9,13 @@
 from numpy.testing import assert_array_equal
 
 
+def asdf_open_memory_mapping_kwarg(memmap: bool) -> dict:
```
braingram, line 12: This package can use the (sometimes confusing) `minversion` from astropy:
https://github.com/astropy/asdf-astropy/blob/dab5b4d73ce3f75e60e5b29e75eb3e811dfd8319/asdf_astropy/converters/transform/core.py#L134

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -128,7 +136,7 @@ def test_no_memmap(tmp_path):
         af.write_to(file_path)
 
     # Update a value in the ASDF file
-    with asdf.open(file_path, mode="rw", copy_arrays=True) as af:
+    with asdf.open(file_path, mode="rw", asdf_open_memory_mapping_kwarg(memmap=False)) as af:
```
braingram, line 139:
```suggestion
    with asdf.open(file_path, mode="rw", **asdf_open_memory_mapping_kwarg(memmap=False)) as af:
```

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -138,7 +146,7 @@ def test_no_memmap(tmp_path):
         assert (af.tree["quantity"] != quantity).any()
         assert (af.tree["quantity"] == new_quantity).all()
 
-    with asdf.open(file_path, mode="rw", copy_arrays=True) as af:
+    with asdf.open(file_path, mode="rw", asdf_open_memory_mapping_kwarg(memmap=False)) as af:
```
braingram, line 149:
```suggestion
    with asdf.open(file_path, mode="rw", **asdf_open_memory_mapping_kwarg(memmap=False)) as af:
```

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -4,9 +4,17 @@
...
 from astropy import units
 from astropy.units import Quantity
+from astropy.utils.introspection import minversion
 from numpy.testing import assert_array_equal
 
 
+def asdf_open_memory_mapping_kwarg(memmap: bool) -> dict:
+    if minversion("asdf", "3.1.0"):
+        return {"memmap": memmap}
+    else :
```
braingram, line 14:
```suggestion
    else:
```

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -4,9 +4,17 @@
 from asdf.testing import helpers
 from astropy import units
 from astropy.units import Quantity
+from astropy.utils.introspection import minversion
 from numpy.testing import assert_array_equal
 
 
+def asdf_open_memory_mapping_kwarg(memmap: bool) -> dict:
```
braingram, line 11: The `bool` annotation here is leading to a pre-commit/ruff failure:
https://results.pre-commit.ci/run/github/271820376/1721652625.YDD78eseT3m3vTW-_PD9kw
https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/
I am ok with either of the following:
- a global ignore for all `FBT` rules.
- removing this type hint

braingram:
@zacharyburnett would you rebase this to (hopefully) allow the oldestdeps to use the oldest versions (now that https://github.com/astropy/asdf-astropy/pull/232 is in).

Also pre-commit is complaining of a `RET505` error.
https://results.pre-commit.ci/run/github/271820376/1721656013.8blo7lrzSCa9pRAmDI-kog

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -4,9 +4,17 @@
...
+    if minversion("asdf", "3.1.0"):
+        return {"memmap": memmap}
+    else:
+        return {"copy_arrays": not memmap}
```
braingram, line 15:
```suggestion
    return {"copy_arrays": not memmap}
```
pre-commit is picky

braingram (approved):
Thanks!

braingram:
Oh, looks like we don't have branch protection rules 😮  Let's wait for the CI to go green before merging.

Issue #231, opened 2024-07-18
oldestdeps is not testing oldest dependencies


braingram:
As revealed in https://github.com/astropy/asdf-astropy/pull/230

The oldestdeps job is not installing the oldest version of dependencies and doesn't appear to run `minimum_dependencies` and is using the latest version of our dependencies:
https://github.com/astropy/asdf-astropy/actions/runs/9993708226/job/27621947749?pr=230#step:10:24

PR #232, opened 2024-07-18
drop py39 try to fix oldestdeps

braingram:
This PR drops a few old versions of:
- python, now >=3.10
- numpy, now >=1.24
- astropy, now >=5.2

This PR also fixes the oldestdeps job which appears to have never worked due to a typos, `oldestdep` instead of `oldestdeps` (and `parallels` instead of `parallel`):
https://github.com/astropy/asdf-astropy/blob/dab5b4d73ce3f75e60e5b29e75eb3e811dfd8319/.github/workflows/ci.yml#L103

When it `oldestdeps` working with this PR and the numpy version was bumped the CI showed failures due to the old astropy (5.0.4). These were addressed by bumping astropy to 5.2.0.

Fixes https://github.com/astropy/asdf-astropy/issues/231

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,9 @@ jobs:
       cache-restore-keys: |
         pip-
       # Any env name which does not start with `pyXY` will use this Python version.
-      default_python: '3.9'
+      default_python: '3.12'
       envs: |
-        - linux: py39-parallel-cov
+        - linux: py310-test-parallel-cov
         - linux: py311-test-devdeps-parallel-cov
         - linux: py312-test-predeps-parallel-cov
       coverage: codecov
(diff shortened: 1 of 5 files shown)


braingram:
I think the coverage drop is due to vestigial code that we can remove since we're dropping older versions.

I'd like to get this in to test https://github.com/astropy/asdf-astropy/pull/230 so will make an issue to track a more thorough search of our code for conditionals we can remove now that we don't support certain versions.

braingram:
@zacharyburnett I can't request your review so I pinged @nden instead. This should allow us to test your PR.

perrygreenfield (approved):
LGTM (with my limited expertise in this). Seems pretty straightforward though.

Issue #238, opened 2024-08-21
Add support for serialization of `astropy.nddata.StdDevUncertainty` and `astropy.nddata.UnknownUncertainty` objects.


ViciousEagle03:
Implement the converter class for the below classes which are subclasses of `astropy.nddata.NDuncertainty`

- Implement converter and design schema for `astropy.nddata.nduncertainty.StdDevUncertainty` .

- Implement converter and design schema for `astropy.nddata.nduncertainty.UnknownUncertainty`.

PR #239, opened 2024-08-21
Add serialization logic for uncertainty objects

ViciousEagle03:
Fixes: #238

This PR adds the support for the serialization of below objects to ASDF

`astropy.nddata.nduncertainty.StdDevUncertainty`
`astropy.nddata.nduncertainty.UnknownUncertainty`


ViciousEagle03:
@braingram, @Cadair, could you guys please review it.

braingram:
Recent commits also added `astropy.nddata.nduncertainty.VarianceUncertainty` support.

asdf_astropy/resources/manifests/astropy-1.2.0.yaml
```diff
@@ -6,15 +6,15 @@ description: |-
   model classes, which are handled by an implementation of the ASDF
   transform extension.
 asdf_standard_requirement:
-  gte: 1.6.0
```
braingram, line 9: Same as https://github.com/astropy/asdf-astropy/pull/246#discussion_r1829957042

braingram:
@Cadair I can't request you as a reviewer. Would you give this a review?

I think this one looks good although given that I padded a couple commits an independent review would be helpful.

asdf_astropy/resources/schemas/nddata/stddevuncertainty-1.0.0.yaml
```diff
@@ -0,0 +1,20 @@
+%YAML 1.1
```
perrygreenfield, line 1: Given the near identical nature of these three schemas, have you considered just using one tag and schema to cover all three cases and making the type of uncertainty an attribute in the schema?

asdf_astropy/converters/nddata/uncertainty.py
```diff
@@ -0,0 +1,45 @@
+class _UncertaintyBaseConverter:
```
perrygreenfield, line 1: I think this could all be done by one converter if you took the suggestion later to handle all this with one tag.

perrygreenfield (commented):
Looks good other than I think it could be simplified.

asdf_astropy/resources/schemas/nddata/stddevuncertainty-1.0.0.yaml
braingram: Thanks. I think using a single tag would require putting some sort of class identifier (either a name, code, or converted name) in the data files (unless I'm missing something) to know what class to use for deserialization. I think there are a few downsides to this.

If we use the class name we're further tying the ASDF representation of the astropy API. If astropy decides to rename a class we would have to retain some mapping of old name to new name. By using a different tag per class we would only have to update the converter to add the new class name to the types (which we would also have to do if we used the other approach).

If we create a mapping of class name to converted class name (for example `StdDevUncertainty` to `stdev`) we will be less dependent on the astropy API (so a class name would only require changing the converter). However if a new uncertainty is added we would need to bump the tag version and start managing a different mapping for each tag version. If we use separate tags and astropy adds a new uncertainty we can just add a new tag for the new uncertainty (and don't have to make any changes to the other schemas and converters).

I think using separate tags allows us to more fully use the existing asdf extension/tag versioning features (albeit with duplication in the schemas).

perrygreenfield: It doesn't have to be an explicit class name, just a descriptive attribute (e.g., `uncertainty_type = unknown`, `variance`, or `stddev`). `from_yaml_tree` then can check the value of this attribute and call the appropriate constructor. That would keep it generic enough for any language library.

braingram: How does [34da76e](https://github.com/astropy/asdf-astropy/pull/239/commits/34da76e132b24c327fb821dd5519fc535668a4cd) look?

I added a [note](https://github.com/astropy/asdf-astropy/pull/239/files#diff-ebc6dbfd9facf1aa378b89c9f88a10cfb98639d249ae083649b99a3096150451R9) that the mapping between class name and "name" in the schema will need to be separately versioned if the schema is updated.

perrygreenfield: Yes, this is exactly what I had in mind. I think it makes it easier to add new uncertainty types as well.

perrygreenfield (approved):
LGTM

Issue #233, opened 2024-07-18
Remove vestigial code for old versions


braingram:
After https://github.com/astropy/asdf-astropy/pull/232 we can go through the codebase and remove code that is no longer reachable.

I think some of these (the asdf-transform-schemas ones) go beyond the deps changed in the linked PR (another reason to have this be a separate PR):
https://github.com/search?q=repo%3Aastropy%2Fasdf-astropy%20minversion&type=code

PR #241, opened 2024-09-24
Cleanup, fix CI, remove out-dated version checks

braingram:
Our `oldestdeps` and `devdeps` [CI](https://github.com/astropy/asdf-astropy/actions/runs/10988582459) are failing on main. This PR fixes the CI by:
- removing scipy dev installation. This is an indirect requirement via astropy (we use 0 scipy API). Their nightly wheels are currently broken and given our limited use it seems safe to me to not require a dev scipy version (since the latest released version works with nightly numpy).
- increasing the asdf minimum version. Something with the very old version of jsonschema pulled in by our minimum asdf wasn't playing nice with the CI. I first attempted to bump this to 2.14 but that failed due to a lack of a jsonschema upper pin. 2.14.4 has an upper pin and allows the `oldestdeps` to now pass. 2.14.4 was released Mar 20, 2023 but given the issues with older dependencies I'm ok with this bump.

I also went through and removed vestigial code only run for versions older than what we now support.
Fixes: https://github.com/astropy/asdf-astropy/issues/233

There are a few other very minor changes that I will call out in comments.

Note that some of the CI cleanup in this PR will require updates to our branch protection rules (for example the required `
test / py312-test-parallel` is no longer run and instead we can require `CI / core / py312-test-parallel-cov`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,8 +38,8 @@ jobs:
       default_python: '3.12'
       envs: |
         - linux: py310-test-parallel-cov
-        - linux: py311-test-devdeps-parallel-cov
-        - linux: py312-test-predeps-parallel-cov
+        - linux: py311-test-parallel-cov
+        - linux: py312-test-parallel-cov
       coverage: codecov
 
   asdf-schemas:
@@ -68,9 +68,6 @@ jobs:
       # Any env name which does not start with `pyXY` will use this Python version.
       default_python: '3.12'
       envs: |
-        - linux: py310-test-parallel
-        - linux: py311-test-parallel
-        - linux: py312-test-parallel
         - macos: py311-test-parallel
         - windows: py311-test-parallel
 
@@ -85,9 +82,9 @@ jobs:
       # Any env name which does not start with `pyXY` will use this Python version.
       default_python: '3.12'
       envs: |
-        - linux: py310-test-devdeps-parallel
-        - linux: py311-test-devdeps-parallel
-        - linux: py311-test-devdeps-numpydev-parallel
+        - linux: py312-test-predeps-parallel
+        - linux: py313-test-devdeps-parallel
+        - linux: py312-test-devdeps-numpydev-parallel
 
   oldest:
     needs: [core, asdf-schemas]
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -38,8 +38,8 @@ jobs:
       default_python: '3.12'
       envs: |
         - linux: py310-test-parallel-cov
-        - linux: py311-test-devdeps-parallel-cov
```
braingram, line 41: Removed devdeps and predeps as those are run as part of `devdeps` below.

.github/workflows/ci.yml
```diff
@@ -68,9 +68,6 @@ jobs:
       # Any env name which does not start with `pyXY` will use this Python version.
       default_python: '3.12'
       envs: |
-        - linux: py310-test-parallel
-        - linux: py311-test-parallel
-        - linux: py312-test-parallel
```
braingram, line 73: These are duplicates of the coverage tests run above.

.github/workflows/ci.yml
```diff
@@ -85,9 +82,9 @@ jobs:
       # Any env name which does not start with `pyXY` will use this Python version.
       default_python: '3.12'
       envs: |
-        - linux: py310-test-devdeps-parallel
-        - linux: py311-test-devdeps-parallel
-        - linux: py311-test-devdeps-numpydev-parallel
+        - linux: py312-test-predeps-parallel
+        - linux: py313-test-devdeps-parallel
```
braingram, line 86: Added python 3.13 testing.

pyproject.toml
```diff
@@ -92,6 +92,7 @@ extend-ignore = [
     "DTZ", # flake8-datetimez
     "TD", # flake8-todos
     "FIX", # flake8-fixme
+    "SIM", # flake8-simplify
```
braingram, line 95: Ignore `SIM` rules. This wanted to single-line some if statements making them less readable.

tox.ini
```diff
@@ -47,7 +46,7 @@ commands_pre=
 commands =
     pytest \
     parallel: --numprocesses auto \
-    cov: --cov-report xml --cov asdf_astropy
+    cov: --cov-report xml --cov asdf_astropy \
```
braingram, line 49: This escape should allow posargs to work with `-cov` environments.

nden (approved):
LGTM

PR #235, opened 2024-07-24
Support serialization of  `astropy.wcs.WCS` objects to ASDF. 

ViciousEagle03:
This PR is a [WIP] and adds the support for serialization of `astropy.wcs.WCS` to ASDF. 
fixes: #234 
Checklist: 

- [x] Add support for the basic WCSes by preserving the `wcs.to_fits()`. 
- [x] Test

PR #246, opened 2024-11-05
Fits wcs

braingram:
This PR builds off of https://github.com/astropy/asdf-astropy/pull/235
and works around an astropy issue where SIP coefficients lose precision: https://github.com/astropy/astropy/issues/17334


asdf_astropy/resources/manifests/astropy-1.2.0.yaml
```diff
@@ -6,15 +6,15 @@ description: |-
   model classes, which are handled by an implementation of the ASDF
   transform extension.
 asdf_standard_requirement:
-  gte: 1.6.0
```
braingram, line 9: Note to reviewers this manifest requires ASDF standard 1.6.0 which has not yet become the stable release. Since this PR adds new features that are compatible with the current stable ASDF standard 1.5.0 I chose to change this manifest (`astropy-1.2.0`) to instead support the stable version (using all the pre 1.6.0 schemas) and adding the new WCS schemas.

I also added `astropy-1.3.0` which support 1.6.0 and also adds the new WCS schemas for that standard version.

So to summarize. Before this PR:
- `astropy-1.2.0` supported ASDF standard 1.6.0 (the unused development version)

After this PR:
- `astropy-1.2.0` supports ASDF standard 1.5.0 and adds the WCS schemas
- `astropy-1.3.0` supports ASDF standard 1.6.0 and adds the WCS schemas

braingram:
@Cadair I can't request you as a reviewer. Would you give this a review?

ViciousEagle03:
> @ViciousEagle03 I put together this PR based on the work you did serializing `astropy.wcs.WCS`. It turns out one of the hang-ups is a bug in astropy (that was a fun one to track down). If you're available to give this a review that would be great! Thanks again for all your work!

This is great! Thanks for cleaning up the code and finding a workaround for the bug. We can finally merge the rest of the PRs dependent on the serialization of `astropy.wcs.WCS` .

asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - type: object
+    properties:
+      wcs:
+        tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+      slices_array:
+        type: array
+        items:
+          - oneOf:
+            - type: integer
+            - type: object
```
Cadair, line 25: Would it be simpler if we always stored a slice object with stop = start + 1?

asdf_astropy/converters/wcs/tests/test_wcs.py
```diff
@@ -0,0 +1,78 @@
...
+    b = rng.uniform(low=-1e-5, high=1e-5, size=(5, 5))
+
+    # Assign SIP coefficients
+    wcs.sip = Sip(a, b, None, None, wcs.wcs.crpix)
+    wcs.wcs.set()
+
+    return wcs
+
+
+def create_tabular_wcs():
```
Cadair, line 40: This is a distortion table, but do we have a test for a `-TAB` wcs?

Cadair (approved):

asdf_astropy/converters/wcs/tests/test_wcs.py
braingram: Any suggestions for an example to try?

If I use the one here:
https://github.com/astropy/astropy/blob/0f80be718107ef797a58ba42025e70b21c454c61/astropy/wcs/tests/conftest.py#L11
it doesn't roundtrip through pickle (or `to_fits`)
```python
import pickle
from astropy.wcs import WCS
from astropy.wcs.tests.helper import SimModelTAB
st = SimModelTAB(nx=150, ny=200)
hdulist = st.hdulist
wcs = WCS(hdulist[0].header, hdulist)
pickle.loads(pickle.dumps(wcs))
```
```
KeyError: "Extension ('WCS-TABLE', 1) not found."
```

braingram: The file in the astropy package data `get_pkg_data_filename("data/tab-time-last-axis.fits", package="astropy.wcs.tests")` also fails to round-trip through pickle (or `to_fits`).

braingram: Oh https://github.com/astropy/astropy/issues/9998 astropy can't produce a `-TAB` wcs so we can't support it here.

asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
braingram: Possibly. Although I think this should retain the format of the `_slices_array`. Which can contain all of these values.

pllim:
tl;dr -- Isn't the whole point of ASDF is to be able to use GWCS so you can ditch FITS WCS?

asdf_astropy/converters/wcs/tests/test_wcs.py
Cadair: Well I guess we should write a test which errors then :laughing:

asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
Cadair: That's fair, let's leave it as is.

braingram:
5f34952271a3273c231bea2dc6c454238332799a adds some tests using the astropy.wcs.tests.data. I wasn't able to find a definitive list of "wcses that should work" and some of the test data is designed to trigger errors. I tried to get most of the files that were not designed to trigger errors.

This did reveal one bug in this PR where setting "pixel_shape" on deserialization resulted in an error for https://github.com/astropy/astropy/blob/7896e82c0b06ea4b3870711e7124569f812df39f/astropy/wcs/tests/data/spectra/orion-velo-1.hdr since the "pixel_shape" is inconsistent with "naxis". I fixed this by removing the "pixel_shape" setting (and looking back I don't see this set anywhere in [__WCS_unpickle__](https://github.com/astropy/astropy/blob/7896e82c0b06ea4b3870711e7124569f812df39f/astropy/wcs/wcs.py#L3584) which was the model for this PR).

perrygreenfield (approved):
LGTM, as much as FITS WCS looks good to me ;-)

PR #253, opened 2025-01-28
Add support for masked quantities, angles, and coordinates

### LINKED ISSUE #202: Support MaskedQuantity
Astropy (v5.0+) has the `Masked` class (https://docs.astropy.org/en/stable/utils/masked/index.html), which can produce `MaskedQuantity`. It would be great if ASDF could serialize Masked objects.

lpsinger:
Fixes #202.


lpsinger:
> I read this as masked stating that the `MaskedQuantity` class exists at `astropy.utils.masked.core.MaskedQuantity` but it doesn't. For asdf to check for instances to handle with the Converter it needs to have access to the class.

That's right. `Masked` is apparently a class factory, and types are dynamically created here: https://github.com/astropy/astropy/blob/v7.0.0/astropy/utils/masked/core.py#L266-L271

The fully qualified name `astropy.utils.masked.core.MaskedQuantity` seems to be totally fictional because it is not added to the module dictionary. Even if it wasn't, I don't see how that fully qualified name would distinguish between instances of the class `Masked(astropy.units.Quantity)` vs `Masked(foo.bar.Quantity)` (a class with a name that is coincidentally the same but comes from a completely different module.

Is there a way to configure a converter to decide whether it is capable of handling instances of a class by calling a function?

braingram:
> Is there a way to configure a converter to decide whether it is capable of handling instances of a class by calling a function?

Not that I'm aware of. The matching of converter to object uses the object type:
https://github.com/asdf-format/asdf/blob/a0898c2d8d5afad0f8cf9a0d60847263d786f9e5/asdf/yamlutil.py#L297
and the assumption is that either:
- asdf has indexed the class (it was provided via `Converter.tags` as is done in this PR at the expense of importing the required modules to make the class at "integration" time)
- asdf has the "path" via which the class can be imported (so asdf can check if the contained module is imported and index the converter as needed)

(see this [comment](https://github.com/asdf-format/asdf/blob/a0898c2d8d5afad0f8cf9a0d60847263d786f9e5/asdf/extension/_manager.py#L64) for more details).

I suspect that more extensive changes to asdf will be needed to support `Masked` since the classes are not publicly importable via astropy (we can't provide a "path" to the class in `Converter.tags`).

lpsinger:
> I suspect that more extensive changes to asdf will be needed to support `Masked` since the classes are not publicly importable via astropy (we can't provide a "path" to the class in `Converter.tags`).

What do you have in mind?

It is turning out to be surprisingly simple to implement `Masked` classes; the significant hurdle is how to register the classes without importing them.

lpsinger:
One option could be to have Astropy add those dynamically-created subclasses to the module dictionary, so that (for example) `Masked(astropy.units.Quantity)` is available as `astropy.utils.masked.Masked_astropy_units_Quantity` (would need some way to transform the fully qualified name of the data class so that it is a valid Python identifier).

braingram:
> > I suspect that more extensive changes to asdf will be needed to support `Masked` since the classes are not publicly importable via astropy (we can't provide a "path" to the class in `Converter.tags`).
> 
> What do you have in mind?

I will try to carve out some time to look into this soon. At the moment I don't have a clear solution to propose.

lpsinger:
> > > I suspect that more extensive changes to asdf will be needed to support `Masked` since the classes are not publicly importable via astropy (we can't provide a "path" to the class in `Converter.tags`).
> > 
> > 
> > What do you have in mind?
> 
> I will try to carve out some time to look into this soon. At the moment I don't have a clear solution to propose.

For a possible workaround, see https://github.com/astropy/astropy/pull/17685.

pllim:
Are you able to run a job here using your branch at https://github.com/astropy/astropy/pull/17685 and see if it is actually green? Thanks! (devdeps is usually the easiest to hack for such things)

lpsinger:
> Are you able to run a job here using your branch at [astropy/astropy#17685](https://github.com/astropy/astropy/pull/17685) and see if it is actually green? Thanks! (devdeps is usually the easiest to hack for such things)

... maybe? See latest commit 🤞

tox.ini
```diff
@@ -35,7 +35,7 @@ extras =
 commands_pre=
     numpydev: pip install -U --pre -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
     devdeps: pip install -U --pre -i https://pypi.anaconda.org/liberfa/simple pyerfa
-    devdeps: pip install -U --pre -i https://pypi.anaconda.org/astropy/simple astropy
+    devdeps: pip install -U --pre -i git+https://github.com/lpsinger/astropy@masked-classes-importable
```
pllim, line 38:
```suggestion
    devdeps: pip install git+https://github.com/lpsinger/astropy@masked-classes-importable
```

pllim: Oh right, your fork won't have the right tags for setuptools_scm. You can make a fake tag on your fork, or otherwise relax any astropy pins here. The former might be easier.

lpsinger:
> Oh right, your fork won't have the right tags for setuptools_scm. You can make a fake tag on your fork, or otherwise relax any astropy pins here. The former might be easier.

I'm not really sure how to to make this work in the CI against a branch of astropy. But the tests I added are passing on my computer.

pllim:
`astropy==0.2.dev37528+g4b8a62975`

Try grab v7.1.dev tag from astropy proper and push that tag to your fork.  Then rerun the jobs.

pllim:
https://github.com/astropy/astropy/releases/tag/v7.1.dev

braingram:
https://github.com/astropy/asdf-astropy/pull/254/commits/dfee9194c90c7e6da17106f4487a77be11a3addb
is one way to appease (parts of) the CI. With that change the tests (added in this PR) are passing with the astropy PR:
https://github.com/astropy/asdf-astropy/actions/runs/13018230750/job/36312572977?pr=254
@lpsinger would you integrate that change with this PR (and drop the tox changes). Not everything will pass (the py310 jobs will fail with astropy dev which doesn't support python 3.10) but hopefully it will illustrate that the astropy changes make this PR possible.

braingram:
Overall this looks good to me. Once the astropy PR is merged we can:
- restore the pyproject.toml astropy version
- update this PR with the appropriate version checks to get things working with all supported (and dev) astropy versions

Thanks again for the PR and adding support for `Masked` classes!

lpsinger:
https://github.com/astropy/astropy/pull/17685 was merged, and it's milestoned for astropy 7.1.0. Currently, asdf-astropy depends on astropy >= 5.2.0. Will it be OK to bump that to 7.1.0?

pllim:
> Will it be OK to bump that to 7.1.0?

Probably not, since this package is used in jwst pipeline. Is it possible to do astropy version check instead?

lpsinger:
> > Will it be OK to bump that to 7.1.0?
> 
> Probably not, since this package is used in jwst pipeline. Is it possible to do astropy version check instead?

I'll see what I can do.

lpsinger:
> > > Will it be OK to bump that to 7.1.0?
> > 
> > 
> > Probably not, since this package is used in jwst pipeline. Is it possible to do astropy version check instead?
> 
> I'll see what I can do.

Done. Take a look.

asdf_astropy/converters/utils.py
```diff
@@ -18,3 +18,26 @@ def import_type(type_name):
...
+
+# FIXME: Remove once we depend on astropy >= 7.1.
+def import_masked_type(class_name):
+    try:
+        return import_type(f"astropy.utils.masked.core.Masked{class_name}")
+    except AttributeError as e:
+        from astropy import __version__
+        from packaging.version import Version
+
+        if Version(__version__) < Version("7.1dev"):
```
pllim, line 41: Maybe move `ASTROPY_LT_7_1` outside of testing only and reuse here?

Also not sure how discoverable is all the notes about FIXME. Maybe use ASTROPY_LT_7_1 as marker in the comment as well?

Thanks!

lpsinger: > Maybe move `ASTROPY_LT_7_1` outside of testing only and reuse here?

Isn't it a no-no to import Astropy at module scope? https://www.asdf-format.org/projects/asdf/en/latest/asdf/extending/converters.html#entry-point-performance-considerations

lpsinger: > Also not sure how discoverable is all the notes about FIXME. Maybe use ASTROPY_LT_7_1 as marker in the comment as well?

Do you mean that you just want that string to appear in all of the FIXME comments, like this?

```python
# FIXME: ASTROPY_LT_7_1: Remove this once we depend on astropy >= 7.1
```

pllim: That works for me. Thanks! Just need something consistent we can `grep` when the time comes. :)

lpsinger: Done. Back to my other question...

> > Maybe move `ASTROPY_LT_7_1` outside of testing only and reuse here?
> 
> Isn't it a no-no to import Astropy at module scope? [https://www.asdf-format.org](https://www.asdf-format.org/projects/asdf/en/latest/asdf/extending/converters.html#entry-point-performance-considerations)

pllim: > Isn't it a no-no to import Astropy at module scope?

Argh... right, the circular import. 😭 

OK then maybe slap `# ASTROPY_LT_7_1` comment there for future `grep` and be done with it? Thanks!

lpsinger: I already put it in the `# FIXME` comment at the top of this function, noting that the entire function should be removed. Is that sufficient?

pllim: Yeah, I missed that earlier. LGTM now. I will resolve this thread. Thanks!

pllim:
oldestdeps job crashed doing coverage or something...

The code LGTM but I'll let the actual maintainer review and approve. Thanks!

lpsinger:
> oldestdeps job crashed doing coverage or something...

The tests are passing now.

braingram:
> > oldestdeps job crashed doing coverage or something...
> 
> The tests are passing now.

Thanks for working on this.

The oldestdeps job is currently segfaulting: https://github.com/astropy/asdf-astropy/actions/runs/14068343837/job/39396519918?pr=253#step:10:202
any thoughts on what's causing that?

I'll take a look at the rest of this PR shortly.

pllim:
> The tests are passing now.

All but one (coverage not included).

asdf_astropy/converters/coordinates/earth_location.py
```diff
@@ -3,10 +3,15 @@
...
-    types = ("astropy.coordinates.earth.EarthLocation",)
+    types = (
+        "astropy.coordinates.earth.EarthLocation",
+        "astropy.utils.masked.core.MaskedEarthLocation",
+    )
 
     def to_yaml_tree(self, obj, tag, ctx):
-        return obj.info._represent_as_dict()
+        result = obj.info._represent_as_dict()
+        result.pop("__class__", None)
```
braingram, line 13: Under what condition was `__class__` being included in the `_represent_as_dict` result?

asdf_astropy/converters/unit/tests/resources/test.asdf
braingram, line 1: Let's find a different way of testing this that doesn't require a test file to be included in the package. Is it sufficient to inline a YAML-only version for the test?

lpsinger:
> The oldestdeps job is currently segfaulting: https://github.com/astropy/asdf-astropy/actions/runs/14068343837/job/39396519918?pr=253#step:10:202
> any thoughts on what's causing that?

No idea. It's been doing that intermittently. Retrying the job seems to fix it usually.

asdf_astropy/converters/utils.py
```diff
@@ -18,3 +18,26 @@ def import_type(type_name):
...
+    def __init__(self, class_name):
+        from astropy import __version__
+
+        super().__init__(
+            f"ASDF support for masked {class_name} objects requires Astropy 7.1 or newer, but you have {__version__}",
+        )
+
+
+# FIXME: ASTROPY_LT_7_1: Remove once we depend on astropy >= 7.1.
+def import_masked_type(class_name):
```
braingram, line 34: Is this only used in `QuantityConverter`? If so, let's just inline this code there.

asdf_astropy/testing/helpers.py
```diff
@@ -2,7 +2,14 @@
 Helpers for testing astropy objects in ASDF files.
 """
 
+import astropy
 import numpy as np
+import pytest
+from packaging.version import Version
+
+ASTROPY_LT_7_1 = Version(astropy.__version__) < Version("7.1.dev")
```
braingram, line 10: Let's move these changes into the tests files since this is importing pytest.

asdf_astropy/converters/coordinates/tests/test_masked.py
```diff
@@ -0,0 +1,74 @@
+import asdf
+import numpy as np
+import pytest
+from astropy import units as u
+from astropy.coordinates import Angle, EarthLocation, Latitude, Longitude, SkyCoord
+from astropy.utils.masked import Masked
+
+from asdf_astropy.testing.helpers import skip_if_astropy_lt_7_1
+
```
braingram, line 9:
```suggestion
pytest.skip(ASTROPY_LT_7_1, allow_module_level=True)
```
Let's just skip this whole file is the astropy version isn't satisfied (we will need to find a suitable place for ASTROPY_LT_7_1) that way we can remove the `skip_if_astropy_lt_7_1` decorator.

lpsinger:
> > The oldestdeps job is currently segfaulting: https://github.com/astropy/asdf-astropy/actions/runs/14068343837/job/39396519918?pr=253#step:10:202
> > any thoughts on what's causing that?
> 
> No idea. It's been doing that intermittently. Retrying the job seems to fix it usually.

I pushed 175a95245f5e68fb886cc8265d470b106156c3fa to try to troubleshoot the segfaults.

asdf_astropy/converters/unit/tests/test_masked.py
```diff
@@ -0,0 +1,41 @@
...
+
+from asdf_astropy.converters.utils import MaskedClassesRequireAstropy71Error
+from asdf_astropy.testing.helpers import only_if_astropy_lt_7_1, skip_if_astropy_lt_7_1
+
+from . import resources
+
+MaskedQuantity = Masked(u.Quantity)
+
+
+@only_if_astropy_lt_7_1
```
braingram, line 17: Let's move this outside this test file and add a module-level skip like mentioned in the other comment. That way we don't need to add the extra decorators.

braingram (commented):
I left an initial set of comments. I will likely have more as this progresses.
One thing I'm not sure about is where to put `ASTROPY_LT_7_1`. The version check will be in a few files so having it in some centralized location makes sense. However all uses currently are in tests so perhaps `asdf_astropy.tests.versions.ASTROPY_LT_7_1`. What do you think?

asdf_astropy/converters/utils.py
lpsinger: It is right now, but I anticipate that we will have more uses as we add masked support to more converters.

asdf_astropy/converters/coordinates/earth_location.py
lpsinger:
```pycon
>>> from astropy import units as u
>>> from astropy.coordinates import EarthLocation
>>> from astropy.utils.masked import Masked
>>> lon = Masked([0, 1] * u.deg)
>>> lat = Masked([2, 3] * u.deg)
>>> el = EarthLocation.from_geodetic(lon, lat)
>>> el.info._represent_as_dict()
{'x': <MaskedQuantity [6374277.59793627, 6368484.28240152] m>, 'y': <MaskedQuantity [     0.        , 111162.30664365] m>, 'z': <MaskedQuantity [221104.54531112, 331574.31534289] m>, 'ellipsoid': 'WGS84', '__class__': 'astropy.coordinates.earth.EarthLocation'}
```

asdf_astropy/converters/coordinates/tests/test_masked.py
lpsinger: I don't think that `pytest.skip` takes a boolean option. You're thinking of `pytest.mark.skipif`. But I did something similar; please take a look.

asdf_astropy/converters/unit/tests/test_masked.py
lpsinger: You mean, a new file, like `test_masked_not_implemented.py`?

asdf_astropy/testing/helpers.py
lpsinger: Done.

asdf_astropy/converters/utils.py
braingram: Thanks. Let's deal with that in the PRs that add those.

asdf_astropy/converters/coordinates/tests/test_masked.py
braingram: Thanks. Let's go with:
```python
if ASTROPY_LT_7_1:
    pytest.skip("These tests require astropy >= 7.1dev", allow_module_level=True)
```
to allow removal of `skip_if_astropy_lt_7_1`.

asdf_astropy/converters/unit/tests/test_masked.py
braingram: `test_quantity.py` seems fine to me something like:
```python
from asdf_astropy.tests.versions import ASTROPY_LT_7_1


@pytest.mark.skipif(not ASTROPY_LT_7_1, reason="MaskedQuantity support was added in astropy 7.1")
def test_masked_quantity_raises():
    import importlib.resources
    from asdf_astropy.converters.utils import MaskedClassesRequireAstropy71Error
    from . import resources

    def test_masked_quantity_raises():
        with pytest.raises(MaskedClassesRequireAstropy71Error):
            with asdf.open(importlib.resources.files(resources) / "test.asdf"):
                pass
```
pending adding `ASTROPY_LT_7_1` to that location (and also ignoring that this is using the test data file to be removed).

.github/workflows/ci.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       # Any env name which does not start with `pyXY` will use this Python version.
       default_python: '3.11'
       envs: |
-        - linux: py311-test-oldestdeps-parallel-cov
+        - linux: py311-test-oldestdeps-cov
```
braingram, line 99: Let's change this back.

braingram (changes requested):

asdf_astropy/tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+import astropy
```
braingram, line 1: Would you put this into a submodule? How about `asdf_astropy.tests.versions` based on the previous suggestions.

asdf_astropy/converters/utils.py
```diff
@@ -18,3 +18,26 @@ def import_type(type_name):
     """
     module_name, class_name = type_name.rsplit(".", 1)
     return getattr(importlib.import_module(module_name), class_name)
+
+
+# FIXME: ASTROPY_LT_7_1: Remove once we depend on astropy >= 7.1.
+class MaskedClassesRequireAstropy71Error(NotImplementedError):
```
braingram, line 24: I don't think this needs a custom exception or the astropy version.

asdf_astropy/converters/coordinates/tests/test_masked.py
lpsinger: Done.

asdf_astropy/converters/unit/tests/test_masked.py
lpsinger: Done.

asdf_astropy/tests/__init__.py
lpsinger: Done.

asdf_astropy/converters/utils.py
lpsinger: This is now just an ordinary `NotImplementedError`. Please take a look.

asdf_astropy/converters/utils.py
lpsinger: If I inline it, then I will have to remove `test_masked_quantity_raises`, and the final `raise` clause here will be missing from code coverage:

```python
# FIXME: ASTROPY_LT_7_1: Remove once we depend on astropy >= 7.1.
def import_masked_type(class_name):
    try:
        return import_type(f"astropy.utils.masked.core.Masked{class_name}")
    except AttributeError as e:
        from astropy import __version__
        from packaging.version import Version

        if Version(__version__) < Version("7.1.dev"):
            msg = (
                f"ASDF support for masked {class_name} objects requires "
                f"Astropy 7.1 or newer, but you have {__version__}"
            )
            raise NotImplementedError(msg) from e
        raise  # <-- this line will be missed
```

.github/workflows/ci.yml
lpsinger: Done.

It's segfaulting when we run the tests serially too. That narrows it down a bit.

asdf_astropy/converters/unit/tests/resources/test.asdf
lpsinger: Done.

lpsinger:
I think that the test coverage should be a lot higher than what codecov.io is reporting. I suspect it's not correctly merging the results from the various jobs. @pllim, any insights?

pllim:
Looks like upload is skipped in devdeps (https://github.com/astropy/asdf-astropy/actions/runs/14071589848/job/39406882171?pr=253)

Maybe you need to add something like this for that job:

https://github.com/astropy/asdf-astropy/blob/0a68c556945ee0a411974f7d4e828eb450297146/.github/workflows/ci.yml#L100C7-L100C24

lpsinger:
> Looks like upload is skipped in devdeps (https://github.com/astropy/asdf-astropy/actions/runs/14071589848/job/39406882171?pr=253)
> 
> Maybe you need to add something like this for that job:
> 
> https://github.com/astropy/asdf-astropy/blob/0a68c556945ee0a411974f7d4e828eb450297146/.github/workflows/ci.yml#L100C7-L100C24

That's better. Now it has 100% coverage.

asdf_astropy/converters/coordinates/earth_location.py
lpsinger: @taldcroft, @mhvk, I most often see your handles associated with `_represent_as_dict` methods. Is it a bug for the `__class__` key to be present in its return value for masked classes?

asdf_astropy/converters/unit/quantity.py
```diff
@@ -39,4 +45,6 @@ def from_yaml_tree(self, node, tag, ctx):
             value = value._make_array()
             dtype = value.dtype
 
-        return Quantity(value, unit=node["unit"], copy=copy, dtype=dtype)
+        class_ = import_masked_type("Quantity") if isinstance(value, np.ma.MaskedArray) else Quantity
```
braingram, line 48:
```suggestion
        if isinstance(value, np.ma.MaskedArray):
            if not minversion("astropy", "7.1.dev"):    # ASTROPY_LT_7_1
                msg = "MaskedQuantity support requires astropy 7.1 or later"
                raise NotImplementedError(msg)
            class_ = Masked(Quantity)
        else:
            class_ = Quantity
```
This requires adding 2 more imports around line 36:
```python
from astropy.utils import minversion
from astropy.utils.masked import Masked
```

asdf_astropy/converters/unit/quantity.py
```diff
@@ -1,6 +1,8 @@
 from asdf.extension import Converter
 from asdf.tags.core.ndarray import NDArrayType
 
+from asdf_astropy.converters.utils import import_masked_type
+
```
braingram, line 5:
```suggestion
```

asdf_astropy/converters/utils.py
braingram: This can be deleted with the other changes.

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -149,3 +151,22 @@ def test_no_memmap(tmp_path):
...
+    data: [1.0, 2.0, null]
+    mask: !core/ndarray-1.0.0
+      data: [false, false, true]
+      datatype: bool8
+      shape: [3]
+    datatype: float64
+    shape: [3]
+"""
+    buff = helpers.yaml_to_asdf(yaml, version="1.5.0")
+    with pytest.raises(NotImplementedError):
```
braingram, line 171:
```suggestion
    with pytest.raises(NotImplementedError, match="MaskedQuantity support requires astropy 7.1 or later"):
```

lpsinger:
I replaced `ASTROPY_LT_7_1` with `ASTROPY_GE_7_1` so that we can reuse the `minversion` function.

asdf_astropy/converters/unit/tests/test_quantity.py
```diff
@@ -149,3 +151,22 @@ def test_no_memmap(tmp_path):
         assert af.tree["quantity"][-1, -1] != new_value
         assert (af.tree["quantity"] != new_quantity).any()
         assert (af.tree["quantity"] == quantity).all()
+
+
+@pytest.mark.skipif(not ASTROPY_GE_7_1, reason="MaskedQuantity support was added in astropy 7.1")
```
braingram, line 156:
```suggestion
@pytest.mark.skipif(ASTROPY_GE_7_1, reason="MaskedQuantity support was added in astropy 7.1")
```

lpsinger: Fixed.

braingram (approved):
Thanks for bringing this across the finish line!

PR #255, opened 2025-01-28
Drop python 3.10, numpy 1.24, update CI

braingram:
Also:
- drops `numpydev` so we no longer separate numpy dev version and other dependency dev testing (all dev versions are tested together)
- fixes dkist CI (same as https://github.com/asdf-format/asdf/commit/a0898c2d8d5afad0f8cf9a0d60847263d786f9e5)


lpsinger:
Oops. I opened https://github.com/astropy/asdf-astropy/pull/256 which partly duplicates this.

braingram:
@lpsinger How does this look to you? Let me know if there are version changes etc that I missed or if you have any questions.

lpsinger (approved):

lpsinger:
> @lpsinger How does this look to you? Let me know if there are version changes etc that I missed or if you have any questions.

Looks great!

PR #263, opened 2025-02-14
remove astropy_init

braingram:


asdf_astropy/__init__.py
```diff
@@ -1,6 +1,3 @@
-# Packages may add whatever they like to this file, but
-# should keep this content at the top.
-# ----------------------------------------------------------------------------
-from ._astropy_init import *  # noqa: F403
+__all__ = ["__version__"]
```
pllim, line 1: Don't think you need this.
```suggestion
```

pllim (approved):
Thanks! Just a minor comment.

PR #265, opened 2025-03-10
Pre commit update

braingram:
This update the pre-commit hooks.

These haven't been updated in some time since the automatic PR was failing since automatic fixes didn't cover all required changes.

This PR changes the default rule selection removing `ALL` and replacing it with the rules that currently apply from `ALL`. This will hopefully allow future pre-commit automatic PRs to automatically pass checks.


pllim (approved):
Your ruff-fu is sure better than mine. Thanks!

PR #276, opened 2025-05-15
Fix two WCS bugs

Cadair:
In trying to polish up sunpy/ndcube#776 I found these two bugs.

### DIFF
--- a/asdf_astropy/converters/wcs/wcs.py
+++ b/asdf_astropy/converters/wcs/wcs.py
@@ -2,7 +2,7 @@ from asdf.extension import Converter
 
 # These attributes don't end up in the hdulist and
 # instead will be stored in "attrs"
-_WCS_ATTRS = ("naxis", "colsel", "keysel", "key", "pixel_bounds")
+_WCS_ATTRS = ("naxis", "_naxis", "colsel", "keysel", "key", "pixel_bounds")
 
 
 class WCSConverter(Converter):
@@ -18,6 +18,8 @@ class WCSConverter(Converter):
         if naxis := attrs.pop("naxis"):
             hdulist[0].header["naxis"] = naxis
 
+        _naxis = attrs.pop("_naxis")
+
         pixel_bounds = attrs.pop("pixel_bounds")
 
         wcs = WCS(hdulist[0].header, fobj=hdulist, **attrs)
@@ -29,6 +31,9 @@ class WCSConverter(Converter):
             wcs.wcs.set()
 
         wcs.pixel_bounds = pixel_bounds
+
+        if _naxis:
+            wcs._naxis = _naxis
         return wcs
 
     def to_yaml_tree(self, wcs, tag, ctx):
--- a/asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
+++ b/asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
@@ -16,7 +16,10 @@ allOf:
   - type: object
     properties:
       wcs:
-        tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+        anyOf:
+          - tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+          - tag: "tag:stsci.edu:gwcs/wcs-*"
+
       slices_array:
         type: array
         items:


asdf_astropy/resources/schemas/wcs/slicedwcs-1.0.0.yaml
```diff
@@ -16,7 +16,10 @@ allOf:
   - type: object
     properties:
       wcs:
-        tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+        anyOf:
+          - tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+          - tag: "tag:stsci.edu:gwcs/wcs-*"
```
Cadair, line 21: I assume adding this to the current schema is fine as it just extends it, but happy to make a new one.

asdf_astropy/converters/wcs/wcs.py
```diff
@@ -18,6 +18,8 @@ def from_yaml_tree(self, node, tag, ctx):
         if naxis := attrs.pop("naxis"):
             hdulist[0].header["naxis"] = naxis
 
+        _naxis = attrs.pop("_naxis")
```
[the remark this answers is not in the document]
Cadair: The reason for adding this was that [`WCS.pixel_shape`](https://github.com/astropy/astropy/blob/main/astropy/wcs/wcsapi/fitswcs.py#L241) wasn't roundtripping properly. I think that this is technically a duplication of the `NAXISn` keys, but you would have to ask a wcs.WCS expert about why it isn't going through the header.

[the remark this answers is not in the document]
braingram: How about we write out (and read in) `pixel_shape` instead of `_naxis`?

Since this wasn't caught by the unit tests would you update those the check for `pixel_shape` roundtripping?

Cadair:
I don't know what to do about these test fails. Some of the test headers we are using from astropy have *some* missing NAXISn keys, one in particular has NAXIS1 but not 2-4. This means that the pixel shape doesn't roundtrip, because astropy sets pixel_shape without validation when reading the header apparently, but not when we explicitly set it using the setter.

Interestingly, this did not happen when we were using the private `_naxis` API because that went through the header. I propose we go back to `_naxis` as I think it's actually the least worst option, and I'll go open some issues on astropy about all this and hopefully we can get it sorted in the future?

braingram:
I opened a PR against this PR:
https://github.com/Cadair/asdf-astropy/pull/1

Let me know what you think.

Cadair:
Thanks @braingram I did consider that option, but I wasn't sure if you wanted to yank all the spectral WCSes. We should definitely open an issue upstream about the invalid headers.

What other issues should we open upstream?

* That `pixel_shape` can be set to an invalid value.
* That `NAXISn` isn't round-tripped through a header.

asdf_astropy/resources/schemas/wcs/slicedwcs-1.1.0.yaml
```diff
@@ -0,0 +1,44 @@
...
+  It manages the slicing and coordinate transformations while preserving
+  the underlying WCS object.
+
+allOf:
+  - type: object
+    properties:
+      wcs:
+        anyOf:
+          - tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+          - tag: "tag:stsci.edu:gwcs/wcs-*"
```
Cadair, line 21: This also needs to support all the ndcube wrappers as well:


```suggestion
          - tag: "tag:stsci.edu:gwcs/wcs-*"
          - tag: "tag:sunpy.org:ndcube/resampledwcs-*"
          - tag: "tag:sunpy.org:ndcube/ndcubesequence-*"
          - tag: "tag:sunpy.org:ndcube/reorderedwcs-*"
          - tag: "tag:sunpy.org:ndcube/compoundwcs-*"
```

Although as I said on #296 (https://github.com/astropy/asdf-astropy/pull/296#discussion_r2336218911) maybe we should just be allowing any object here?

braingram: Would you help me understand how is ndcubesequence a wcs? I'm not familiar with the class.

Cadair: Whoops, copy paste error.

braingram:
> Thanks @braingram I did consider that option, but I wasn't sure if you wanted to yank all the spectral WCSes.

I'm ok dropping them given their odd behavior after loading. Maybe we can re-add them in the future if they're updated.

> We should definitely open an issue upstream about the invalid headers.
> 
> What other issues should we open upstream?
> 
>     * That `pixel_shape` can be set to an invalid value.
> 
>     * That `NAXISn` isn't round-tripped through a header.

I'm not familiar enough with astropy WCS to know what's considered an issue.

asdf_astropy/resources/schemas/wcs/slicedwcs-1.1.0.yaml
braingram: I'm leaning towards adding each of these (with 1.x wildcards: `wcs-1.*`). There's probably a somewhat academic discussion to be had about leaking in-memory API structure into the serialization format since as you mentioned here the goal is to allow anything that is API compliant (which doesn't jive with a constraint on data structure).
I'm also ok with allowing any object here since as noted the intended constraint is on API and not on structure.
I'll leave this one up to you. I'm ok with either the suggestion you added or allowing any object.

asdf_astropy/resources/schemas/wcs/slicedwcs-1.1.0.yaml
```diff
@@ -0,0 +1,44 @@
...
+  to the WCS, allowing certain pixel and world dimensions to be retained or
+  dropped.
+  It manages the slicing and coordinate transformations while preserving
+  the underlying WCS object.
+
+allOf:
+  - type: object
+    properties:
+      wcs:
+        anyOf:
```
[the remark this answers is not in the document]
braingram: With https://github.com/astropy/asdf-astropy/pull/276/commits/47bd317e915b708eeb6b6a52834a0bd667ed67a9 adding "any object" this now needs to be `anyOf`

Cadair:
@braingram can we merge this?

PR #281, opened 2025-06-05
Ignore UnitsWarnings

braingram:
This PR ignores UnitsWarnings when serializing and deserializing units.

### DIFF
--- a/asdf_astropy/converters/unit/tests/test_unit.py
+++ b/asdf_astropy/converters/unit/tests/test_unit.py
@@ -18,16 +18,20 @@ def vounit_compatible(unit):
 
 
 def create_vounits():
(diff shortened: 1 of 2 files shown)


asdf_astropy/converters/unit/tests/test_unit.py
```diff
@@ -18,16 +18,20 @@ def vounit_compatible(unit):
 
 
 def create_vounits():
-    return {u for u in list(units.__dict__.values()) if isinstance(u, units.UnitBase) and vounit_compatible(u)}
+    return sorted(
```
braingram, line 21: This is unrelated but while testing this I noticed that the test order was non-deterministic.

PR #296, opened 2025-09-10
Add support for HighLevelWCSWrapper

Cadair:
This is used quite heavily in ndcube.

I added use of the lazy fixtures package here, the test changes for SlicedLowLevelWCS are kind of related to #276 in that it makes it easy to also test with gwcs.

### DIFF
--- a/asdf_astropy/converters/wcs/highlevelwrapper.py
+++ b/asdf_astropy/converters/wcs/highlevelwrapper.py
@@ -3,7 +3,7 @@ from asdf.extension import Converter
 
 class HighLevelWCSWrapperConverter(Converter):
     tags = ("tag:astropy.org:astropy/wcs/highlevelwcswrapper-*",)
-    types = ("astropy.wcs.wcsapi.HighLevelWCSWrapper",)
+    types = ("astropy.wcs.wcsapi.high_level_wcs_wrapper.HighLevelWCSWrapper",)
 
     def from_yaml_tree(self, node, tag, ctx):
         from astropy.wcs.wcsapi import HighLevelWCSWrapper
--- a/asdf_astropy/converters/wcs/highlevelwrapper.py
+++ b/asdf_astropy/converters/wcs/highlevelwrapper.py
@@ -1,0 +1,14 @@
+from asdf.extension import Converter
+
+
+class HighLevelWCSWrapperConverter(Converter):
+    tags = ("tag:astropy.org:astropy/wcs/highlevelwcswrapper-*",)
+    types = ("astropy.wcs.wcsapi.HighLevelWCSWrapper",)
+
+    def from_yaml_tree(self, node, tag, ctx):
+        from astropy.wcs.wcsapi import HighLevelWCSWrapper
+
+        return HighLevelWCSWrapper(node["wcs"])
+
+    def to_yaml_tree(self, hlwcs, tag, ctx):
+        return {"wcs": hlwcs.low_level_wcs}
--- a/asdf_astropy/converters/wcs/tests/test_highlevelwcswrapper.py
+++ b/asdf_astropy/converters/wcs/tests/test_highlevelwcswrapper.py
@@ -1,0 +1,36 @@
+import asdf
+import pytest
+import gwcs
+from astropy.wcs import WCS
+from astropy.wcs.wcsapi import HighLevelWCSWrapper
+from astropy.wcs.wcsapi.wrappers.sliced_wcs import SlicedLowLevelWCS
+
+from asdf_astropy.testing.helpers import assert_wcs_equal
+
+from pytest_lazy_fixtures import lf
+
+
+@pytest.fixture
+def wrapped_wcses(all_4d_wcses):
+    return HighLevelWCSWrapper(all_4d_wcses)
+
+
+@pytest.fixture
+def wrapped_sliced_wcs(astropy_wcs_4d):
+    return HighLevelWCSWrapper(SlicedLowLevelWCS(astropy_wcs_4d, 1))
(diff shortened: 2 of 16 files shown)


asdf_astropy/resources/schemas/wcs/highlevelwcswrapper-1.0.0.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+description: >-
+    Wrapper class that can take any astropy.wcs.wcsapi.BaseLowLevelWCS
+    object and expose the high-level WCS API.
+
+allOf:
+  - type: object
+    properties:
+      wcs:
+        anyOf:
```
Cadair, line 16: It would be really good to have a way of saying "Any APE-14 compliant WCS". I am kind of tempted to allow any object inside these WCS wrapper classes as we can't possibly enumerate all of them?

Now I think about it #276 has the same problem with the changes to the SlicedLowLeveWCS schema, we should be allowing all the NDCube wrappers there as well.

tox.ini
```diff
@@ -29,6 +29,8 @@ deps =
     oldestdeps: minimum_dependencies
 
     parallel: pytest-xdist
+
+    gwcs: gwcs>=0.22.0
```
Cadair, line 33: The oldest version of gwcs with the `examples` module needs asdf 3.3 so we can't have it in the mix with the rest of the oldestdeps build. Doing this makes it optional and hopefully makes it easier to test old asdfs etc.

asdf_astropy/resources/schemas/wcs/highlevelwcswrapper-1.0.0.yaml
```diff
@@ -0,0 +1,26 @@
...
+          - tag: "tag:astropy.org:astropy/wcs/wcs-1*"
+          - tag: "tag:stsci.edu:gwcs/wcs-*"
+          - tag: "tag:astropy.org:astropy/wcs/slicedwcs-*"
+          - tag: "tag:sunpy.org:ndcube/resampledwcs-*"
+          - tag: "tag:sunpy.org:ndcube/ndcubesequence-*"
```
Cadair, line 21:
```suggestion
```

Cadair:
Is this oldestdeps fail a asdf standard < 1.6 error? What's the best solution?

asdf_astropy/converters/wcs/highlevelwrapper.py
```diff
@@ -0,0 +1,14 @@
+from asdf.extension import Converter
+
+
+class HighLevelWCSWrapperConverter(Converter):
+    tags = ("tag:astropy.org:astropy/wcs/highlevelwcswrapper-*",)
+    types = ("astropy.wcs.wcsapi.HighLevelWCSWrapper",)
```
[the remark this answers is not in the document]
Cadair: Pretty sure we added it with the original wcsapi. It's older than SlicedLowLevelWCS iirc.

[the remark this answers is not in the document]
Cadair: Yeah I guess it's defined in that file and pulled up a level. I can add the full path if you want?

tox.ini
braingram: Thanks. It sounds like there are other reasons to bump to 3.3 (2.15 is ancient and https://github.com/astropy/asdf-astropy/pull/296#discussion_r2336658866 will allow for just the public class path).

I'll open a PR to bump that pin to see if it triggers any other changes.

asdf_astropy/converters/wcs/tests/test_highlevelwcswrapper.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.mark.parametrize("hl_wcs", [lf("wrapped_wcses"), lf("wrapped_sliced_wcs")])
+def test_hllwcs_serialization(hl_wcs, tmp_path):
+    file_path = tmp_path / "test_highlevelwcswrapper.asdf"
+    with asdf.AsdfFile() as af:
+        af["hl_wcs"] = hl_wcs
+        af.write_to(file_path)
+
+    ll_wcs = hl_wcs._low_level_wcs
```
braingram, line 29: Is there a public way to get the low level wcs?

Cadair:
gwcs.WCS and astropy.wcs.WCS both implement the high and low level APIs. Things such as the wrapper classes only implement the low level api so the high level wrapper gives you a way to get a high level object.

Yes probably we don't strictly need to test all combinations of things if it feels lile overkill 😆

braingram:
> gwcs.WCS and astropy.wcs.WCS both implement the high and low level APIs. Things such as the wrapper classes only implement the low level api so the high level wrapper gives you a way to get a high level object.
> 
> Yes probably we don't strictly need to test all combinations of things if it feels lile overkill 😆

Thanks for the clarification. That sounds good to me. Would you update this PR dropping gwcs as a test dependency and making the required test changes?

braingram (changes requested):
Thanks for putting this together. There have been a few asdf-astropy releases since this PR was started so some new files will need to be created and changes to existing files reverted.

braingram (approved):
Thanks for adding this new feature. @Cadair I can't request you as a reviewer for your PR :) but would you take a look since I pushed a few commits. If this looks good to you I'm happy to have it merged.

Is this functionality something you need in a new release soon or can it sit around until the next release (no plan for that at the moment).

braingram:
0.11.0 is up on pypi with these changes:
https://pypi.org/project/asdf-astropy/0.11.0/

Thanks again!

PR #300, opened 2025-10-06
Avoid using the private `EarthLocation._get_site_registry()` in tests

eerovaher:
The development dependencies CI job is failing because the tests here rely on private `astropy` code that has been changed: https://github.com/astropy/asdf-astropy/actions/runs/18271678280/job/52015325062#step:10:201 Luckily the replacement code (in the first commit) that avoids private API is even simpler and also backwards-compatible.

It doesn't look to me like `test_earthlocation_site` is worth keeping at all because it just duplicates one of the parametrizations of `test_serialization`, so I removed it. The removal is in a separate commit so that it would be simple to undo in case it does serve some purpose.


braingram (approved):
Thanks! LGTM.

PR #322, opened 2026-04-14
update workflow for trusted publishing

### LINKED ISSUE #321: REL/SEC: enabling trusted publishing for PyPI
[A recent supply chain attack](https://blog.pypi.org/posts/2026-04-02-incident-report-litellm-telnyx-supply-chain-attack/) highlighted the inherent risks of using long-lived tokens to publish packages to PyPI, which was the only option for automated deployment [until Trusted Publishing was introduced in 2023](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/). `OpenAstronomy`'s publishing workflows were initially designed around the token-based strategy, and [recently added support for using trusted publishing instead](https://github.com/OpenAstronomy/github-actions-workflows/releases/tag/v2.4.0).

I'm reporting to all projects I find still relying on the token-based strategy, and strongly recommend this project switches to trusted publishing. I'm happy to answer any questions maintainers may have about the process, and I note that most of the work requires administration rights both to the repository and the PyPI project.

The PyPI token stored as a secret associated to this repo should be revoked (i.e., removed) in order to minimize the risk of it being stolen. I advise taking this action immediately (this takes only a minute) even if maintainers don't have time to set up trusted publishing just yet.

braingram:
Closes https://github.com/astropy/asdf-astropy/issues/321

Update publish workflow to use trusted publishing.

I've configured pypi to trust this workflow.

Once this is in we can make a dev/alpha release to test the process.

https://github.com/braingram/asdf-astropy/actions/runs/24423796233/job/71353151823
is an attempt to test with testpypi (which I did not configure to trust this workflow, does testpypi support this?). I'm sharing that test since it shows that the general process is working where:
- the prior build/publish workflow step builds and uploads an artifact but does not attempt to publish
- the new publish workflow downloads the artifacts and attempts the upload

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -8,14 +8,30 @@ permissions:
   contents: read
 
 jobs:
-  build-n-publish:
+  build:
     permissions:
       contents: none
     if: (github.event_name == 'release')
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@e97344095b099e1d729fe97429078c9975921d8a  # v2.6.2
     with:
-      upload_to_pypi: true
+      upload_to_pypi: false
+      save_artifacts: true
       test_extras: test
       test_command: pytest --pyargs asdf_astropy
-    secrets:
-      pypi_token: ${{ secrets.PYPI_PASSWORD }}
+
+  upload:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    needs: [build]
+    environment: pypi
+    permissions:
+      id-token: write
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          merge-multiple: true
+          pattern: dist-*
+          path: dist
+      - name: Upload to PyPI
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0


braingram:
@neutrinoceros any chance you can take a look at this? Any advice for testing this beyond merging then making a test dev release? Thanks!

.github/workflows/publish-to-pypi.yml
```diff
@@ -8,14 +8,30 @@ permissions:
...
+      id-token: write
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          merge-multiple: true
+          pattern: dist-*
+          path: dist
+      - name: Upload to PyPI
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
```
neutrinoceros, line 37: you *could* run an end-to-end (ish) test without polluting the real registry with
```suggestion
        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
        with:
            repository-url: https://test.pypi.org/legacy/
```
but be warned that it might take more effort to setup than it's worth:
- you'd need a separate account for the test registry
- you'd need to setup trusted publishing there too, so in the end you'd be testing just the GH side, not PyPI

overall, I suggest doing this only if you're really uncertain that you'll remember having done this next time you need an actual release, because it would complicate your debugging process if anything goes wrong then. And if it *does* go wrong, don't hesitate to ping me again, I'd gladly give a hand !

neutrinoceros, line 37: You could also allow the workflow to run *partially* (skipping the upload step) on PRs

```suggestion
        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
        if: github.event_name == 'release'
```

you'd need to add something like this to the trigger list (top level `on:`)
```yml
  pull_request:
    paths:
    - .github/workflows/publish-to-pypi.yml
    - pyproject.toml
```

neutrinoceros (commented):
this looks good to me as is, though I understand why you're worried about the inability to test it end to end before an actual release is made. Here are a couple comments and suggestions for testing *most* of it without making a test-only release. Let me know if you have other questions !

.github/workflows/publish-to-pypi.yml
braingram: Thanks! I'll take your advice on this being more effort than it's worth :)

braingram: Thanks for the suggestion! There's a `build` step as part of the PR triggered CI:
https://github.com/astropy/asdf-astropy/blob/0fb5a282dee066649023ea27d4de318443728e5a/.github/workflows/ci.yml#L105
I'll consider removing that, updating this workflow and changing the branch protection rules but hold off on doing so for a separate PR.

braingram:
Thanks for taking a look at this!

Cadair:
If you run the action on a PR or branch it will work all the way up to the upload step itself where it will error on a permissions error.

Cadair (approved):

